PR #8, opened 2016-09-30
Don't install OpenSSL on appveyor

alexcrichton:
Also don't need config on OSX as we're using SecureTransport there.


frewsxcv:
FYI, presumably relevant build fails on Appveyor.

sfackler:
We use OpenSSL as a client for server tests

alexcrichton:
Aha, nevermind!

PR #9, opened 2016-09-30
Set some defaults based on libcurl

alexcrichton:
* Enable `SSL_OP_ALL` by default which apparently enables a number of useful bug
  workarounds (claimed by libcurl)
* Frob a few bits here afterwards to disable some insecure options.
* Use the cipher list from libcurl to enable/disable ciphers


src/imp/openssl.rs
```diff
@@ -112,6 +112,18 @@ impl<S> From<ErrorStack> for HandshakeError<S> {
 fn ctx() -> Result<SslContext, Error> {
     let mut ctx = try!(SslContext::new(SslMethod::Sslv23));
     try!(ctx.set_default_verify_paths());
+
+    // optiosn to enable and cipher list lifted from libcurl
```
frewsxcv: Nit: "options"

src/imp/openssl.rs
```diff
@@ -112,6 +112,18 @@ impl<S> From<ErrorStack> for HandshakeError<S> {
...
     try!(ctx.set_default_verify_paths());
+
+    // options to enable and cipher list lifted from libcurl
+    let mut opts = ssl::SSL_OP_ALL;
+    opts |= ssl::SSL_OP_NO_TICKET;
+    opts |= ssl::SSL_OP_NO_COMPRESSION;
+    opts &= !ssl::SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG;
+    opts &= !ssl::SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS;
+    opts |= ssl::SSL_OP_NO_SSLV2;
+    opts |= ssl::SSL_OP_NO_SSLV3;
```
sfackler: We may want to force disable sslv2 and sslv3 on the other implementations for uniformity. This seems reasonable though.

alexcrichton: I think that's a good point yeah. I've pushed a commit to disable it on Windows, but calling [`set_protocol_version_min`](http://sfackler.github.io/rust-security-framework/doc/v0.1.8/security_framework/secure_transport/struct.SslContext.html#method.set_protocol_version_min) on OSX will require some more finesse because it's only defined on 10.8+

sfackler: I think I'm probably okay with not supporting 10.7. We can also use the [`SSLSetProtocolVersionEnabled`](https://developer.apple.com/reference/security/1503754-sslsetprotocolversionenabled) function. It's now deprecated in favor of `SSLSetProtocolVersionMin` and `SSLSetProtocolVersionMax` but it still exists IIRC.

alexcrichton:
Ok, @sfackler I've updated where both schannel and security-framework disable SSL v2/v3

sfackler:
Thanks!

PR #11, opened 2016-10-31
Restructure API

sfackler:
Use high level OpenSSL wrappers for configuration logic as well.

cc @alexcrichton Not totally sold on the naming here, but I can't really think of anything better.


alexcrichton:
Seems reasonable to me!

PR #14, opened 2016-11-23
Derive Clone interface for Pkcs12

taheris:
Because `TlsConnectorBuilder::identity` consumes its `Pkcs12` argument, this derives the Clone trait to avoid re-parsing.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -68,6 +68,7 @@ impl From<ErrorStack> for Error {
     }
 }
 
+#[derive(Clone)]
 pub struct Pkcs12(pkcs12::ParsedPkcs12);
 
 impl Pkcs12 {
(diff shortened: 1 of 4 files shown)


sfackler:
You'll need to pipe the change through the openssl crate looks like.

Out of curiosity, why were you needing to construct many connectors with the same identity? A single connector can create as many TLS streams as you want.

taheris:
Yep, not as straightforward as this :) I was hoping to re-use the same connector (with a bound identity) in multiple Hyper clients, similar to how they are created in the Reqwest library [here](https://github.com/seanmonstar/reqwest/blob/a580af2a226351d57a0a8151fc3f523039339c40/src/client.rs#L73).

sfackler:
An easy way of doing that would be to stick it in an `Arc`. You may need to define your own `SslClient` implementation though.

Issue #13, opened 2016-11-21
Add ability to disable verification


seanmonstar:
I know, I know, it's horrible. It is also a reality that many people deal with this, and so it would be useful to add a way to disable verification. It can and probably should be a terrifying long name, `very_insecure_disable_verification` or what-have-you.

cc https://github.com/seanmonstar/reqwest/issues/15

sfackler:
How much of verification? Just hostname or the whole deal?

seanmonstar:
I assume certificate verification. Besides the host name verification, the rest is to verify from a set of trusted CAs that the certificate is signed by one of them? I've never needed it myself, but the case I imagine is people have a self-signed certificate. In that case, hostname still makes sense, but checking the CAs would cause problems.

alex:
I'd start with making it easier to specify an _alternate_ validator, e.g. provide a `PinnedValidator` which people can put their self-signed cert in to trust that. Try to find ways to support every semi-reasonable usecase that people want to disable stuff for. And only when that finally fails add the degenerate `VeryInsecureNoValidationValidator`

sfackler:
Definitely agree on that approach.

Adding trusted CAs is in some ways orthogonal to the validators, and is just blocked on adding support to schannel-rs. Unfortunately, this is not really a thing Windows wants you to do (trusted CAs must either be registered in the system trust root, or be present in a Certificate Trust List signed by something trusted by your system trust root) but it should be doable.

Philipp91:
Maybe the `VeryInsecureNoValidationValidator` could be hidden behind a cargo feature called `make_me_vulnerable_to_mitm` or something like that. The concrete security implications should be made clear -- not only the general fact that it's less secure.

sfackler:
Hostname verification is in #15.

emk:
I'm currently dealing with a production emergency involving `native-tls`, PostgreSQL and Citus's hosted database product. This is occurring in my company's first production Rust app. :-( (Happily, nobody's trying to use it this afternoon, but I need to fix it quickly nonetheless.)

The problem is that PostgreSQL SSL cert validation is still fairly rare in non-Rust libraries, and plenty of PostgreSQL servers have self-signed or invalidly-signed certs. It's hard to debug because PostgreSQL uses [a custom protocol to enable SSL](https://www.postgresql.org/docs/9.3/static/protocol-flow.html#AEN99579). I'd be debugging faster if I weren't sleep deprived from work travel. :-(

[...]

PR #15, opened 2016-11-29
Allow connection without hostname verification

sfackler:
cc #13

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -375,6 +375,21 @@ impl TlsConnector {
         let s = try!(self.0.connect(domain, stream));
         Ok(TlsStream(s))
     }
+
+    /// Like `connect`, but does not validate the server's domain name against its certificate.
+    ///
+    /// # Warning
+    ///
+    /// You should think very carefully before you use this method. If hostname verification is not
+    /// used, *any* valid certificate for *any* site will be trusted for use from any other. This
+    /// introduces a significant vulnerability to man-in-the-middle attacks.
+    pub fn danger_connect_without_providing_domain_for_certificate_verification_and_server_name_indication<S>(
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -375,6 +375,21 @@ impl TlsConnector {
...
     }
+
+    /// Like `connect`, but does not validate the server's domain name against its certificate.
+    ///
+    /// # Warning
+    ///
+    /// You should think very carefully before you use this method. If hostname verification is not
+    /// used, *any* valid certificate for *any* site will be trusted for use from any other. This
+    /// introduces a significant vulnerability to man-in-the-middle attacks.
+    pub fn danger_connect_without_providing_domain_for_certificate_verification_and_server_name_indication<S>(
```
frewsxcv: 👌

Philipp91:
If this is merged before https://github.com/sfackler/rust-native-tls/issues/3 is solved or another sensible way to use self-signed certificates (https://github.com/sfackler/rust-native-tls/issues/13#issuecomment-262103881) is available, people will start using `danger_connect_without_providing_domain_for_certificate_verification_and_server_name_indication` despite its name, because it's the only way.

That's of course their own risk. But during the time when this is the only solution, it might make its way onto platforms like StackOverflow as the "accepted solution", as was the case with Android here:
http://stackoverflow.com/a/7266768
Later, clear warnings were added to this answer, but they weren't present before. Research has shown that such code did in fact end up in production code of numerous Android apps (though it remains unclear if it was indeed copied from StackOverflow) and made them vulnerable.

Otherwise, I think the function name is well chosen and will attract enough attention.

sfackler:
Self signed certificates will still fail to validate with this method - all that changes here is that we don't check the SAN/CN of the leaf.

Philipp91:
Ah right.

PR #16, opened 2016-12-15
Expose a few more details of underlying backends

alexcrichton:
This commit updates the various backends to export some more information, where
all of these changes were motivated in getting the tokio-tls test suite running
again. (it's just a bunch of smoke tests really). A summary of the changes are:

* Add an `ErrorExt` trait to each backend to lower a `native_tls::Error` to the
  backend-specific error. This allows inspection of errors codes and such if
  necessary.

* Add `TlsConnectorBuilderExt` to the `security_framework` backend, with one
  function called `anchor_certificates`. This function transitively calls the
  security framework crate's `anchor_certificates` function to set a custom
  trusted certificate for a connection. Along the way this required using the
  `ClientBuilder` type which in turn required changes to work with async I/O.

The changes to the security framework side of things forces this crate to depend
on the changes in sfackler/rust-security-framework#23, so it shouldn't be merged
before that's updated and released!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,8 +9,8 @@ documentation = "https://docs.rs/native-tls/0.1.0/native_tls"
 readme = "README.md"
 
 [target.'cfg(target_os = "macos")'.dependencies]
-security-framework = { version = "0.1.9", features = ["OSX_10_8"] }
-security-framework-sys = "0.1.9"
+security-framework = { version = "0.1.10", features = ["OSX_10_8"] }
+security-framework-sys = "0.1.10"
 tempdir = "0.3"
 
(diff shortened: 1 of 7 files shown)


sfackler:
Security Framework released.

alexcrichton:
Updated to those changes

sfackler:
We should eventually switch the native-tls side APIs over to version min/version max as well.

compressed:
@sfackler is it possible to get a minor release with this change? It would help for testing on mac without having to use native-tls as a git dependency.

sfackler:
Oops, done!

compressed:
Awesome, thanks so much!

PR #18, opened 2017-01-16
Import .p12 files without creating a temp keychain

### LINKED ISSUE #17: Don't create a temporary keychain to decode Pkcs12
Keychains, other than the login one automatically become locked after a set period of time by default.  If the binary is still running, the user will be presented with a password dialogue that they didn't expect for a keychain they didn't remember creating.  It's possible to decode Pkcs12 without interacting with the system keychain at all using SecPKCS12Import.

BlameOmar:
Fixes #17

### DIFF
--- a/src/imp/security_framework.rs
+++ b/src/imp/security_framework.rs
@@ -80,8 +80,16 @@ impl Pkcs12 {
         let mut imported_identities = options.import(buf)?;
         imported_identities.truncate(1);
         let imported_identity = imported_identities.pop().unwrap();
+
+        // FIXME: Compare the certificates for equality using CFEqual
+        let identity_cert = imported_identity.identity.certificate()?.to_der();
         
-        Ok(Pkcs12{identity: imported_identity.identity, chain: imported_identity.cert_chain})
+        Ok(Pkcs12{
+            identity: imported_identity.identity,
+            chain: imported_identity.cert_chain
+                .into_iter()		
+                .filter(|c| c.to_der() != identity_cert)		
+                .collect(),})
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/imp/security_framework.rs
```diff
@@ -78,34 +74,14 @@ pub struct Pkcs12 {
...
-                .collect(),
-        })
+        let mut options = Pkcs12ImportOptions::new();
+        options.passphrase(pass);
+
+        let mut imported_identities = options.import(buf)?;
+        imported_identities.truncate(1);
+        let imported_identity = imported_identities.pop().unwrap();
+        
+        Ok(Pkcs12{identity: imported_identity.identity, chain: imported_identity.cert_chain})
```
sfackler: I believe we'll want to continue filtering the leaf certificate out of the chain.

sfackler:
I was under the impression that you couldn't load `SecIdentity` values without a keychain, but if that's not the case then great!

src/imp/security_framework.rs
```diff
@@ -78,34 +74,22 @@ pub struct Pkcs12 {
...
-            // FIXME possibly use the chain from the trust result instead?
-            chain: import.cert_chain
-                .into_iter()
-                .filter(|c| c.to_der() != identity_cert)
-                .collect(),
-        })
+        let mut options = Pkcs12ImportOptions::new();
+        options.passphrase(pass);
+
+        let mut imported_identities = options.import(buf)?;
```
sfackler: You'll need to use `try!` rather than `?` since we support pre-1.13.

sfackler:
Thanks!

PR #21, opened 2017-02-13
client CA replacement test

bluejekyll:
Not ready for merge. Initial test to validate CA replacement and the client TLS Connector


bluejekyll:
I'll try and fix this up for Cert pinning instead of custom CA, as it looks like there is no method on macOS to replace a CA right now... right?

sfackler:
It may not be exposed, but the underlying library does support append vs replace so it's an easy API addition on the security-framework side.

bluejekyll:
I'm working on a few other changes closing this for now.

Issue #3, opened 2016-05-03
Allow trusted certificates to be registered


sfackler:
This is straighforward for OpenSSL and Secure Transport, but I'm not sure how to do it with SChannel. We'll need to support both the addition of certs to the existing set as well as replacing the trusted set entirely.

Getting the cert out of PEM into a `CERT_CONTEXT` looks to be doable via `CryptStringToBinary` and `CertCreateCertificateContext`, but I still need to figure it out from there.

cc @retep998 - do you know how SChannel would do this?

steffengy:
@sfackler 
I think one would need to implement manual certificate verification to perform checks like these,
since currently verification is handled by the default windows trust provider.
(there's also a TODO for that in the code)

https://msdn.microsoft.com/en-us/library/windows/desktop/aa378740(v=vs.85).aspx

sfackler:
Windows does things in a very different way - you create a Certificate Trust List which needs to be signed by something with a trusted root. Unless we can force people to install a custom root cert (not reasonable), we may have to manually handle this. The OSCrypto library manually checks the root certificate in the trust chain against its extra roots: https://github.com/wbond/oscrypto/blob/master/oscrypto/_win/tls.py#L591

PR #22, opened 2017-02-22
Add examples from README to examples/

Byron:
That way, they are easy to compile for interested users and
run automatically as part of the test-suite.

### DIFF
--- a/examples/google-connect.rs
+++ b/examples/google-connect.rs
@@ -1,0 +1,17 @@
+extern crate native_tls;
+
+use native_tls::TlsConnector;
+use std::io::{Read, Write};
+use std::net::TcpStream;
+
+fn main() {
+    let connector = TlsConnector::builder().unwrap().build().unwrap();
+
+    let stream = TcpStream::connect("google.com:443").unwrap();
+    let mut stream = connector.connect("google.com", stream).unwrap();
+
+    stream.write_all(b"GET / HTTP/1.0\r\n\r\n").unwrap();
+    let mut res = vec![];
+    stream.read_to_end(&mut res).unwrap();
+    println!("{}", String::from_utf8_lossy(&res));
+}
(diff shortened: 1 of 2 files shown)


sfackler:
Looks like these lines need to be bumped: https://github.com/sfackler/rust-native-tls/blob/master/.travis.yml#L6 https://github.com/sfackler/rust-native-tls/blob/master/.travis.yml#L10

https://travis-ci.org/sfackler/rust-native-tls/jobs/204093322

Byron:
@sfackler Thanks for the hints!
On an unrelated note: do you think it's possible to allow untrusted hosts, like those which are self-signed for the purpose of local testing, using this library?

sfackler:
It is not yet - the functionality needs to be added to schannel-rs first. It's definitely something I do want to support though!

Byron:
Perfect! Is  there a tracking issue I could subscribe to? I am trying to connect to a cassandra instance via TLS using `tokio-cassandra`, and am investigating libraries that support this. `rustls` would be been my favorite, but unfortunately its ciphers are too new :).

sfackler:
#3

Byron:
@sfackler - could you restart the travis jobs? It seems they failed for different reasons.

sfackler:
Restarted Travis - The OpenSSL version will need to be updated in the appveyor config.

Byron:
Thank you - the issue seems to be reproducible though.

@sfackler I am afraid I don't know what is going on here. The only change that might have caused this breakage on travis is to remove the v1.9 restriction for Rust. Appveyor needs a different version, as you say, but I don't know why this branch needs it while master apparently still works.
Looking into this it seems [the last build](https://travis-ci.org/sfackler/rust-native-tls/builds/192430849) ran successfully two months ago!
If you restarted that one, I would hope that similar issues crop up. Those could then be fixed in master, and I just rebase this PR onto it - et voilà.

What do you think :)?

sfackler:
You should be able to rebase now

Byron:
@sfackler Sorry for the great delay! The rebase was performed, and this branch looks like it can be merged now.

sfackler:
Thanks!

PR #23, opened 2017-03-04
WIP: Client auth tests and Custom acceptor CA

bluejekyll:
This has some client auth tests. It also adds custom association for CAs on the server side.


Cargo.toml
```diff
@@ -20,3 +20,11 @@ openssl = "0.9"
 
 [dev-dependencies]
 openssl = { version = "0.9", features = ["v102", "v110"] }
+
+[replace]
+"openssl:0.9.7" = { git = "https://github.com/sfackler/rust-openssl.git", features = ["v102", "v110"] }
```
sfackler: Why is this necessary?

bluejekyll: It's temporary, I don't plan on that being merged. It's necessary until the next release of openssl b/c of my use of the pkcs12 builder.

bluejekyll:
@sfackler since you're looking at this: I've got a problem in getting this to pass. Two issues really:

1) on macOS: I can not get the server side of the connection to accept the client cert. I'm not sure what I'm doing wrong there.

2) on Linux and OpenSSL: there is a hang on client auth. I noticed this in my TRust-DNS work, and started trying to narrow down the cause. I need to throw it into a debugger...

sfackler:
There's a test for client auth in the security-framework repo: https://github.com/sfackler/rust-security-framework/blob/master/security-framework/src/os/macos/secure_transport.rs#L462

bluejekyll:
Hm, can you explain:

https://github.com/sfackler/rust-security-framework/blob/master/security-framework/src/os/macos/secure_transport.rs#L478

```rust
            let stream = p!(listener.accept()).0;

            match ctx.handshake(stream) {
                Ok(_) => panic!("unexpected success"),
                Err(HandshakeError::Failure(_)) => {}
                Err(err) => panic!("unexpected error {:?}", err),
            }
```

Shouldn't that `Failure` be `Interrupted`?

sfackler:
Hmm, looking back at that test that does seem strange... I'm not really sure why that would be expected to fail.

bluejekyll:
The error I'm getting on the server side test on macOS is

`'Failure(Error { code: -9807, message: "invalid certificate chain" })'`

So I was going to start investigating the server side registration of the cert to see if there's an issue there, as the same cert construction works in all other contexts.

sfackler:
I wonder if the configured anchor certs aren't taken into account? It may require breaking on peer certs and doing it manually?

bluejekyll:
I think we might be missing `errSSLServerAuthCompleted` in this block:

https://github.com/sfackler/rust-security-framework/blob/master/security-framework/src/secure_transport.rs#L929

based on this doc: https://developer.apple.com/reference/security/1400161-sslhandshake?language=objc

I'll see if I can find that code.

bluejekyll:
@sfackler , see recent change, this doesn't validate the cert chain... we'll need to figure that out, but it does properly pause on the client and then resume to complete the connection.

bluejekyll:
What do you think of providing a single function abstraction over the three different implementations to handle this case?

It would be something like:

```rust
fn handle_client_auth(cert: X509Type) -> Result<()>;
```

The X509Type would need to be a wrapper type of the specific implementation across the underlying libraries. The complexity here, is that we'll need to establish custom certificate chains for validating the client. I think the cert chain and how it is validated will be somewhat domain specific, but it's probably hairy enough that there should be some standard interfaces for certificate validation. Do you have any thoughts in this area? I imagine these would be custom SslContexts...

bluejekyll:
@anowell this is the PR I mentioned at RustConf I was working on a while back for some mTLS stuff. Do you think it's out of date/unnecessary at this point?

anowell:
Thanks, @bluejekyll. I haven't had a chance to dig deep enough into this change yet, but before I crash for the night I thought I'd point you at [how I'm currently reading client key/cert from a kubernetes config](https://github.com/anowell/kubeclient-rs/blob/db7430acea4259d80758bca4d0207309c48721c6/src/config.rs#L78-L97) and then [turning that into a Pkcs12](https://github.com/anowell/kubeclient-rs/blob/db7430acea4259d80758bca4d0207309c48721c6/src/clients/low_level.rs#L42-L47) that I plumb through [a fork of `reqwest`] to native-tls. Of course issue #27 has some discussion around proposed changes to that API.

PR #25, opened 2017-03-05
pkcs12 load test

### LINKED ISSUE #24: Loading pkcs12 der formatted bytes fails on macOS
This change:

https://github.com/sfackler/rust-native-tls/commit/94a150379d42b9ee6b2d2dbbadde02cde88e5159

To get rid of the temp file, appears to have caused an error in my builds. Rolling back this change fixes my issues. I'm running into the unwrap() on 

https://github.com/sfackler/rust-native-tls/commit/94a150379d42b9ee6b2d2dbbadde02cde88e5159#diff-9cef9dd6c1ba7275f2ad10014dd7697fR82

I was not running into this error before the change.

FYI: @BlameOmar

bluejekyll:
A test for validating the loading of a dynamically generated Cert. This helped expose a problem with macOS always adding new keys to the default keychain, see discussion in #24

This does not fix #24, but is useful for it's validation (if you comment out the serialial_number in the cert).

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -1,4 +1,9 @@
+#[cfg(target_os = "macos")]
+extern crate security_framework;
+
 use openssl::ssl::{self, SslMethod, SslConnectorBuilder};
+use openssl::pkcs12::Pkcs12 as OpenSSLPkcs12;
+use openssl::x509::X509;
 use std::io::{Read, Write};
 use std::net::{TcpStream, TcpListener};
 use std::thread;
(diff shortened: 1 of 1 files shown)


sfackler:
We already have some server-side tests that load pkcs12 archives, so I'm not sure this is strictly necessary, and I'd like to minimize openssl use in tests when possible to hopefully be able to get rid of it soon when we support custom trust roots.

bluejekyll:
Sure, that makes sense.

What is your stance on dynamically generated certificates, etc.? Do you want static files for all tests? I'm asking since I may have some I'm adding to my other PR...

sfackler:
I would advocate for generating them once and dumping them in `tests/`.

Issue #4, opened 2016-08-24
Support cipher suite selection


sfackler:
OpenSSL's the most flexible here, and SChannel is the least. The cross-platform solution will probably look something like an SChannel style enum of algorithms that will be unioned:

```rust
pub enum Algorithm {
    Sha256,
    Aes,
    Dhe,
    Rsa,
    ...
}
```

If you then call
```rust
client_builder.supported_algorithms(&[Algorithm::Rsa, Algorithm::Aes, Algorithm::Sha256, Algorithm::Dhe])
```
would mean all ciphers that use RSA or DHE for key exchange, AES as a bulk cipher, *and* SHA256 as a hash algorithm. That'd translate in SChannel to
```rust
cred_builder.supported_algorithms(&[Algorithm::Rsa, Algorithm::Aes, Algorithm::Sha256. Algorithm::Dhe])
```
and in OpenSSL to (note that we have to generate the cartesian product of key exchange, bulk cipher and hash algorithms)
```rust
ctx.set_cipher_suite("RSA+AES+SHA256:DHE+AES+SHA256:@STRENGTH")
```
and in Secure Transport to a somewhat complicated dance where we load supported ciphers and filter them through the provided algorithms:
```rust
let supported = ctx.supported_ciphers();
let enabled = supported.into_iter().filter(|c| suite_supported(c, algorithms)).collect();
ctx.set_enabled_ciphers(&enabled);
```

One interesting question is what to do if no algorithm in a category is specified (e.g. just `&[Algorithm::Aes, Algorithm::Dhe]` is passed). Should that mean that no suites match and no connections will work, or that *all* algorithms of that type can be used?

An alternative design would be to have separate enums for each algorithm type:
```
client_builder.supported_algorithms(&[KeyExchange::Rsa, KeyExchange::Dhe], &[BulkCipher::Aes], &[Hash::Sha256]);
```
It would be a bit more clear, but more verbose as well.

sfackler:
We'll also probably want to disable unauthenticated and anonymous suites entirely.

cyang1:
I'm interested in this. I took a stab at the `security-framework` and `schannel-rs` implementations, and the public API in https://github.com/cyang1/rust-native-tls/commit/a82bfbe245b76d0aa0d44366f157cb4d6ced3f93. I think the OpenSSL implementation would be simple as you mentioned above.

I don't think my change is necessarily production ready yet, as it's not necessarily maintainable in its current state.

[...]

PR #26, opened 2017-03-06
WIP: add support for rustls

aidanhs:
Putting rustls inside native-tls is motivated by issues like https://github.com/tokio-rs/tokio-tls/issues/16 and https://github.com/seanmonstar/reqwest/pull/63 - if the Rust ecosystem can just depend on one tls provider crate with a common interface and swappable backends, that'd be great (indeed, that's what native-tls offers right now with a hardcoded choice per platform).

This PR is an initial attempt at adding rustls support to native-tls, implementing most of the support for tls clients by re-using code from the license-compatible [hyper-rustls](https://github.com/ctz/hyper-rustls/blob/f5207b1/src/lib.rs). Both client tests (`connect_google`, `connect_bad_hostname`) now pass for me. You can try it out with `RUSTFLAGS='--cfg nonnative_tls' cargo test`.

The largest piece of work to do for the server tests is getting pkcs#12 support (which I've started a [discussion about](https://github.com/briansmith/webpki/issues/41)), but I also want to get some early thoughts on what this will look like to an end-user. One question is how to expose the option to use the rustls backend, as there are a few requirements:

 - it's not ideal to require users to use `RUSTFLAGS="--cfg blah"` (though I'm perfectly happy with it for my own use!)
 - libraries depending on this crate ideally should be blind to the backend (though the `*Ext` traits make that a little tricky) and so they should not generally specify which backend they want - that should be the role of the top level crate being built to specify (if it wants to)
 - you'd like the backends to be mutually exclusive by default, but not forcibly so - it isn't an error to have native-tls with multiple backends (it could rank backends and if not instructed specifically to use one, choose the highest ranked), it's just a bit...odd

I don't know there's a good solution for 'soft' mutual exclusion. [Hard mutual exclusion](https://github.com/rust-lang/cargo/issues/2980) might actually be fine - I'm not sure there's any reason to allow multiple tls backends in the same program (and I don't think this has to cover every possible use-case ever anyway).

So let's say the backends are mutually exclusive and libraries don't get to choose - how do you pick one? I wonder if [automatic features](https://github.com/rust-lang/rfcs/pull/1787) could be leveraged(/misused) here (I've not read them in depth) - I think the natural way you'd combine them with mutual exclusion is a ranking i.e. we detect openssl and rustls, but only one is allowed and so we choose openssl since that's ranked higher, for backwards compatibility if nothing else (in fact, even without mutual exclusion, ranking would be doable inside native-tls). And now the top level can choose the crate that automatically implements their preferred native-tls backend.

Anyway, there are some thoughts, some of which may be crazy :) As I say, I'm totally happy if (when this PR is ready) rustls is merged as an 'unstable' feature only accessible through `RUSTFLAGS`, but I wanted to get the ball rolling on the discussion so you're aware.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,17 @@ security-framework-sys = "0.1.10"
 [target.'cfg(target_os = "windows")'.dependencies]
 schannel = "0.1.1"
 
-[target.'cfg(not(any(target_os = "windows", target_os = "macos")))'.dependencies]
+[target.'cfg(not(any(target_os = "windows", target_os = "macos", nonnative_tls)))'.dependencies]
 openssl = "0.9"
+
+[target.'cfg(all(not(any(target_os = "macos", target_os = "windows")), nonnative_tls))'.dependencies]
+rustls = "0.5"
+webpki-roots = "0.7.0"
 
 [dev-dependencies]
 openssl = { version = "0.9", features = ["v102", "v110"] }
+
+# additional, on top of openssl above
+[target.'cfg(nonnative_tls)'.dev-dependencies]
+rustls = "0.5"
+webpki-roots = "0.7.0"
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -7,4 +7,5 @@ pub mod security_framework;
 pub mod schannel;
 
 #[cfg(not(any(target_os = "macos", target_os = "windows")))]
-pub mod openssl;
+pub mod rustls;
+//pub mod openssl;
--- a/src/backend/rustls.rs
+++ b/src/backend/rustls.rs
@@ -1,0 +1,4 @@
+//! Rustls-specific functionality.
+
+//pub use imp::{TlsConnectorBuilderExt, TlsAcceptorBuilderExt, TlsStreamExt};
+//pub use imp::ErrorExt;
--- a/src/imp/rustls.rs
+++ b/src/imp/rustls.rs
@@ -1,0 +1,386 @@
+extern crate rustls;
+extern crate webpki_roots;
+
+use std::error;
+use std::fmt;
+use std::io;
+use std::marker::PhantomData;
+use std::net::Shutdown;
+use std::sync::Arc;
+
+use self::rustls::Session;
+
+use Protocol;
+
+pub enum Error {
+    IO(io::Error),
+    Tls(rustls::TLSError),
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        match *self {
+            Error::IO(ref e) =>  error::Error::description(e),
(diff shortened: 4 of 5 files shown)


aidanhs:
@ctz ping just so you're aware of this as the author of rustls :)

sfackler:
Thanks for the detailed writeup! I've been a bit conflicted over this exact issue in the past.

First, on the technical side, I'm imagining that if we did this that there'd just be a `force-rustls` feature, which would disable the normal backend in favor of rustls.

However, I'm not sure this matches up with this library's objectives. This crate is designed to specifically target the "I don't particularly care what I'm using but I just want the common case to work with minimal effort on my part" use case. If you want to use RusTLS in particular, you should use RusTLS, just as you should use OpenSSL if you want OpenSSL in particular. An abstract provider interface would be cool - there's some prior art in Hyper that could be yanked out somewhere, but this crate would just be one of many consumers of that interface IMO.

aidanhs:
> However, I'm not sure this matches up with this library's objectives. This crate is designed to specifically target the "I don't particularly care what I'm using but I just want the common case to work with minimal effort on my part" use case.

I had a think about this and I do see your point. However, it might also be worth thinking about use cases - assuming there was a 'tls-interface' crate which (somehow) picked a default tls implementation but also permitted you to override with rustls or something else, when *would* you choose to use native-tls directly? Or, to restate the question, when would you want to forbid any implementation except the platform default?

If positioned like that, tls-interface and native-tls effectively become one entity - there isn't much benefit separating them apart from conceptual purity. Not that that's necessarily a bad thing.

sfackler:
This style of crate is the wrong way to define a "global" abstraction, IMO. The various libraries just don't expose enough of a unified interface for it to be sufficiently powerful. A more limited setup like what's used by [hyper](https://docs.rs/hyper/0.10.5/hyper/net/trait.SslClient.html) or [postgres](https://docs.rs/postgres/0.14.0/postgres/tls/trait.TlsHandshake.html) is the right direction to go, IMO.

aidanhs:
I'm not sure I understand what you mean by "powerful" - the 'crait' (crate-trait) can be very limited in scope so that it covers the bare essentials necessary to perform some task (native-tls being the obvious example of exposing some minimum to do tls).

It seems that I'm effectively talking about virtual packages (https://www.debian.org/doc/debian-policy/ch-relationships.html#s-virtual, http://askubuntu.com/questions/738980/how-does-apt-choose-the-specific-package-to-install-in-order-to-provide-a-virtua) as they're called in apt, though with some niceties to integrate better with cargo/the rust ecosystem. I may raise this idea on the cargo repo once I've thought this through a bit more.

sfackler:
You can't even select cipher suites with native-tls. See #4 for why even an operation like that is arbitrarily hard.

PR #30, opened 2017-03-22
use security-framework for iOS

king6cong:


sfackler:
I'm a bit hesitant to land this until we have some integration testing - I think it should be possible to target the iOS simulator on the mac builders?

king6cong:
@sfackler I'm currently working on a network api client SDK in rust for one app that will be available on multiple platforms(including iOS/Android). TLS support on iOS/Android is so important that I can't go any further without it. (HTTPS api is really fundamental). 

I wonder whether TLS for iOS/Android is ready by now ?

king6cong:
So is https://github.com/tokio-rs/tokio-tls/pull/22

sfackler:
Sure, I definitely understand the desire for this to work on iOS - I'm just looking for Travis to actually test to make sure it is in fact working, and will continue to do so over time.

king6cong:
Thanks 👍

azdlowry:
This doesn't build for me where target_os = "ios" as security_framework doesn't have keychain on ios.

king6cong:
@azdlowry yeah, see https://github.com/sfackler/rust-native-tls/issues/44

azdlowry:
This should be unnecessary now as the changes we're included in #50.

PR #35, opened 2017-05-12
Make acceptors and connectors cloneable

azdlowry:
Building a TlsAcceptor or TlsConnector can be non trivial, so cloning allows easy reuse without having to rebuild from scratch.


sfackler:
Looks good, thanks!

PR #37, opened 2017-05-18
Allow initialization for openssl extension builders

### LINKED ISSUE #36: Allow Backend Customization
Let's say I've created an openssl [`SslAcceptor`](https://docs.rs/openssl/0.9.12/openssl/ssl/struct.SslAcceptor.html) because I want customizations (initialize from pem format, use mozilla_modern/intermediate, etc). I still want to use this crate's [`TlsAcceptor`](https://docs.rs/native-tls/0.1.2/native_tls/struct.TlsAcceptor.html) because this abstraction is used in higher level crates like [tokio-tls](https://github.com/tokio-rs/tokio-tls) -- and it's a good abstraction. The only problem is that it looks like `rust_native_tls::imp::openssl::TlsAcceptor`, which as far as I know is what I'm looking for, isn't exposed to users.

Is there a way to inject customizations depending on the backend?

nickbabcock:
Based on our conversation in #36. Please let me know if this is not what you intended or if this PR needs improvement.

Closes #36

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -202,6 +202,9 @@ impl TlsConnector {
 
 /// OpenSSL-specific extensions to `TlsConnectorBuilder`.
 pub trait TlsConnectorBuilderExt {
+    /// Initialize `TlsAcceptorBuilderExt` from an `SslAcceptorBuilder`.
+    fn from_openssl(builder: SslConnectorBuilder) -> Self;
+
     /// Returns a shared reference to the inner `SslConnectorBuilder`.
     fn builder(&self) -> &SslConnectorBuilder;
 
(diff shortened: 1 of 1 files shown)


src/imp/openssl.rs
```diff
@@ -202,6 +202,9 @@ impl TlsConnector {
 
 /// OpenSSL-specific extensions to `TlsConnectorBuilder`.
 pub trait TlsConnectorBuilderExt {
+    /// Initialize `TlsAcceptorBuilderExt` from an `SslAcceptorBuilder`.
+    fn create(SslConnectorBuilder) -> Self;
```
sfackler: Can you use named parameters in the trait definitions? Can we call this `from_openssl` or something like that?

nickbabcock: yes, I'll make the change. Also I wanted to ask if you had a preference of `create` vs `new` for the function name.

nickbabcock:
Updated PR.

src/imp/openssl.rs
sfackler: Sorry, I meant renaming the method to `from_openssl`, not the argument.

nickbabcock:
Ok, updated again 😊

sfackler:
Sorry, this fell off of my radar!

PR #40, opened 2017-06-21
update schannel

sstecko:


sfackler:
This kind of bump isn't required in general. `"0.1.3"` means `">= 0.1.0, < 0.2.0"`. You should be able to just run `cargo update` locally and pick up the new schannel.

sstecko:
Understood! I do update it in my next PR ( https://github.com/sfackler/rust-native-tls/pull/41 ) because it relies on a breaking change.

PR #42, opened 2017-06-22
specify verification callbacks for schannel

sstecko:
adds a trait in the connectorBuilder that allows users to specify a
verification callback to use in the event that a new connection's
validation fails

### DIFF
--- a/src/imp/schannel.rs
+++ b/src/imp/schannel.rs
@@ -313,18 +313,17 @@ impl<S: io::Read + io::Write> io::Write for TlsStream<S> {
 }
 
 // SChannel-specific verification callback for validating failed certificate checks
-pub trait TlsConnectorBuilderCallbackExt {
+pub trait TlsConnectorBuilderExt {
     /// Callback function that determines whether certificate validation failures that occur when establishing connections
     /// should be overridden. The closure should return Ok if the validation failure is to be overridden.
-    fn callback<F>(&mut self, callback: F) ->  Result<(), Error> 
+    fn verify_callback<F>(&mut self, callback: F) 
         where F: Fn(tls_stream::CertValidationResult) -> io::Result<()> + 'static + Send + Sync;
 }
 
-impl TlsConnectorBuilderCallbackExt for ::TlsConnectorBuilder {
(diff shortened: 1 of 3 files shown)


src/imp/schannel.rs
```diff
@@ -305,6 +312,22 @@ impl<S: io::Read + io::Write> io::Write for TlsStream<S> {
     }
 }
 
+// SChannel-specific verification callback for validating failed certificate checks
+pub trait TlsConnectorBuilderCallbackExt {
```
sfackler: Let's just call this `TlsConnectorBuilderExt` since it may grow to include more than this method.

src/imp/schannel.rs
```diff
@@ -305,6 +312,22 @@ impl<S: io::Read + io::Write> io::Write for TlsStream<S> {
...
+    /// Callback function that determines whether certificate validation failures that occur when establishing connections
+    /// should be overridden. The closure should return Ok if the validation failure is to be overridden.
+    fn callback<F>(&mut self, callback: F) ->  Result<(), Error>
```
sfackler: Can you rename this to `verify_callback` to match the method we're delegating to?

sfackler: No need for it to return an error since it can't fail.

steffengy:
Just a random thought, that's in my oppinion worth thinking about:

Wouldn't it be better to expose a generic `ValidationCallback` (non-schannel) specific here?
This could have a `raw_callback` (similar to `raw_stream`) which would return
a `schannel::CertValidationCallback` for schannel 
(AFAIK rust-openssl also already has such a callback?).

Additionally one could abstract `CertChain`/`Certificate` later on top of it,
which would allow a unified callback interface in the long term.

Just a thought though, @sfackler thoughts?

sfackler:
@steffengy Oh yeah, that's definitely the right approach for most users. We're just doing this first because it's a good thing to have around in any case, and creating the high level interface will require some somewhat nontrivial API work. These low-level accessors are here for things that the high-level API doesn't yet support, and to give people more power than the high-level APIs offer when they do exist.

PR #43, opened 2017-06-23
Enable conversion from internal errors to Error

jethrogb:
This allows making a `native_tls::Error` from e.g. a `openssl::error::ErrorStack`. Previously you'd have to do `.into().into()`, but type inference wasn't smart enough to figure this out.

Note this works because of the blanket `impl Into<T> for T`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,9 +155,9 @@ impl fmt::Debug for Error {
     }
 }
 
-impl From<imp::Error> for Error {
-    fn from(err: imp::Error) -> Error {
-        Error(err)
+impl<T: Into<imp::Error>> From<T> for Error {
+    fn from(err: T) -> Error {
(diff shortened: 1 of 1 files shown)


sfackler:
I'd like people to have to opt into backend specific behavior - can we instead do this by adding a constructor method to an `ErrorExt` trait for each backend?

sfackler:
See here for an example: https://github.com/sfackler/rust-native-tls/pull/37

jethrogb:
I agree with the sentiment but that's a lot of extra and duplicate code for little benefit. You effectively can't use backend-specific types without going through one of the existing traits. This PR just makes it so that when you're using that functionality, you don't have to jump through extra hoops for error handling.

sfackler:
Hmm, I think that seems reasonable.

PR #45, opened 2017-06-28
upgrade security-framework to 0.1.14

king6cong:


sfackler:
I don't believe this is necessary - an `0.1.13` constraint allows `0.1.14`, unless we this doesn't work on `0.1.13`?

PR #50, opened 2017-07-07
Changes to fix ios issues and add CI for ios platforms

azdlowry:
Remove keychain when on ios platform, this requires the latest master from https://github.com/sfackler/rust-security-framework.

Uses https://github.com/SSheldon/rust-test-ios to build and run tests on ios simulator.


azdlowry:
Seems some upstream components have stopped building on ios.

src/test.rs
```diff
@@ -14,239 +14,237 @@ macro_rules! p {
...
-
-    p!(socket.write_all(b"GET / HTTP/1.0\r\n\r\n"));
-    let mut result = vec![];
-    p!(socket.read_to_end(&mut result));
-
-    println!("{}", String::from_utf8_lossy(&result));
-    assert!(result.starts_with(b"HTTP/1.0"));
-    assert!(result.ends_with(b"</HTML>\r\n") || result.ends_with(b"</html>"));
-}
+mod tests {
```
sfackler: Why is there an extra nested module here?

.travis.yml
```diff
@@ -2,14 +2,32 @@ language: rust
...
 matrix:
   include:
   - os: osx
-    rust: 1.13.0
+  - os: osx
+    env: TARGET=x86_64-apple-ios RUST_BACKTRACE=1 RUST_TEST_THREADS=1
 env:
 - RUST_BACKTRACE=1 RUST_TEST_THREADS=1
+before_script:
+- if [ ! -z "$TARGET" ]; then
```
sfackler: What does `$TARGET` represent if it isn't the architecture we're testing? Why are we installing all of these?

src/test.rs
azdlowry: rust_test_ios requires the tests to be indented, and this seemed like the most reliable way. I tried to remove the requirement but it's a substantial piece of work.

.travis.yml
azdlowry: Yes, it's not a good name. I have changed it to better describe what it represents. Rust_test_ios requires all of these targets as it's a hard coded list.

src/test.rs
sfackler: Can you stick a note here saying that's why?

azdlowry:
The incoming changes are quite significant and I have no way of testing (or even building) them locally. There might be a lot of "commit and hope" changes while I figure it out.

azdlowry:
All tests are passed.

sfackler:
Thanks!

PR #52, opened 2017-08-04
[RFC] Support for exclusive cert pinning

jchien14:
Sending this structure out for comments before I proceed with implementation on Windows/Linux


sfackler:
The pinning logic seems reasonable, but I'm not sure about the algorithm stuff.

hinaria:
i'd like to consider renaming `disable_built_in_certs()` to something else that better represents our intent.

`rust-native-tls` does not have any built-in certificates. instead, we want to use an empty store with no roots. if you agree, may i suggest a name like `empty_root_certificates()`?

Issue #34, opened 2017-05-11
Parallelism causes server unit tests to occasionally flake on OSX


jchien14:
Pretty stumped here..

If I add an additional unit test on a fresh checkout of master, I start getting two test failures:
```
---- test::server stdout ----
	thread 'test::server' panicked at 'Failure(Error { code: -9806, message: "connection closed via error" })', src/test.rs:76
stack backtrace:
   1:        0x109c13cbc - std::sys::imp::backtrace::tracing::imp::write::hb7ff6d5abd996ef4
   2:        0x109c16f9e - std::panicking::default_hook::{{closure}}::h53094adb345230b0
   3:        0x109c16b43 - std::panicking::default_hook::h5acf019c6eaf66af
   4:        0x109c17457 - std::panicking::rust_panic_with_hook::h8943f907023b7818
   5:        0x109c172b4 - std::panicking::begin_panic::h946820df60a04b85
   6:        0x109c17222 - std::panicking::begin_panic_fmt::h5face8a00b51b583
   7:        0x109b94c66 - native_tls::test::server::h3847131659cdb976
   8:        0x109bc70ce - <F as test::FnBox<T>>::call_box::h5c3bbb3b38d2930a
   9:        0x109c183ba - __rust_maybe_catch_panic
  10:        0x109bbb5c8 - std::panicking::try::do_call::h988206980d2b27fc
  11:        0x109c183ba - __rust_maybe_catch_panic
  12:        0x109bc230e - <F as alloc::boxed::FnBox<A>>::call_box::h2654ce5ce2ab7b8e
  13:        0x109c16664 - std::sys::imp::thread::Thread::new::thread_start::h4ad0b40513420e9c
  14:     0x7fff94e8aaaa - _pthread_body
  15:     0x7fff94e8a9f6 - _pthread_start

---- test::server_tls11_only stdout ----
	thread 'test::server_tls11_only' panicked at 'Failure(Error { code: -9806, message: "connection closed via error" })', src/test.rs:116
```

Removing the test causes things to pass again. Reproduced on 10.12.3.

This is the test I added for posterity:
```#[test]
fn connect_bad_hostname_ignored2() {
    let builder = p!(TlsConnector::builder());
    let builder = p!(builder.build());
    let s = p!(TcpStream::connect("google.com:443"));
    builder.danger_connect_without_providing_domain_for_certificate_verification_and_server_name_indication(s).unwrap();
}
``` 
[...]

PR #54, opened 2017-08-18
Store TempDir for the lifetime of a program

blackbeam:
This solves #34.

Might be an overkill but i can't find another solution.
I've tried to store tempdir within Pkcs12 and even tried to store it in a wrapper for SecKeychain but with no luck.

### DIFF
--- a/src/imp/security_framework.rs
+++ b/src/imp/security_framework.rs
@@ -16,12 +16,13 @@ use std::fmt;
 use std::io;
 use std::error;
 use std::sync::Mutex;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Once, ONCE_INIT};
 
 use Protocol;
 
+static SET_AT_EXIT: Once = ONCE_INIT;
+
 lazy_static! {
-    static ref AT_EXIT_SET: AtomicBool = AtomicBool::new(false);
     static ref TEMP_DIRS: Mutex<Vec<TempDir>> = Mutex::new(vec![]);
 }
 
@@ -110,13 +111,13 @@ impl Pkcs12 {
         // FIXME: Compare the certificates for equality using CFEqual
         let identity_cert = try!(import.identity.certificate()).to_der();
 
-        if !AT_EXIT_SET.swap(true, Ordering::Relaxed) {
+        SET_AT_EXIT.call_once(|| {
             extern "C" fn atexit() {
                 TEMP_DIRS.lock().unwrap().clear();
             }
 
             unsafe { libc::atexit(atexit); }
-        }
+        });
 
         TEMP_DIRS.lock().unwrap().push(dir);
 
--- a/src/imp/security_framework.rs
+++ b/src/imp/security_framework.rs
@@ -1,3 +1,4 @@
+extern crate libc;
 extern crate security_framework;
 extern crate security_framework_sys;
 extern crate tempdir;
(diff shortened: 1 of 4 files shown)


src/imp/security_framework.rs
```diff
@@ -14,9 +15,16 @@ use self::tempdir::TempDir;
...
 use std::io;
 use std::error;
+use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use Protocol;
 
+lazy_static! {
+    static ref AT_EXIT_SET: AtomicBool = AtomicBool::new(false);
+    static ref TEMP_DIRS: Mutex<Vec<TempDir>> = Mutex::new(vec![]);
```
sfackler: Can we treat this as a map from the raw DER data to the directory? That way we don't end up with tons of directories when the same identity is imported multiple times.

src/imp/security_framework.rs
```diff
@@ -101,6 +110,16 @@ impl Pkcs12 {
         // FIXME: Compare the certificates for equality using CFEqual
         let identity_cert = try!(import.identity.certificate()).to_der();
 
+        if !AT_EXIT_SET.swap(true, Ordering::Relaxed) {
```
sfackler: You can use std::sync::Once to do this.

blackbeam:
If SecKeystore isn't cached, then you'll get "keychain with this name already exists" error, so now it's also cached.

src/imp/security_framework.rs
```diff
@@ -79,17 +94,27 @@ pub struct Pkcs12 {
...
+            Occupied(entry) => entry.get().0.clone(),
         };
 
-        let mut keychain = try!(keychain::CreateOptions::new().password(pass).create(
-            dir.path().join("tmp.keychain"),
-        ));
-        // disable lock on sleep and timeouts
-        try!(keychain.set_settings(&KeychainSettings::new()));
-
         let mut imports = try!(
```
sfackler: Will things behave okay if we import the same identity into the same keychain twice?

blackbeam:
>Why do we need a custom hash function here?

It's faster for long keys. It's just a habit and i'll remove it if you say so.

> Will things behave okay if we import the same identity into the same keychain twice?

I've tried to import same der multiple times and it looks like the number of identities in a keychain stays the same (1 identity as shown by ItemSearchOptions).
Anyway there is no need to import same der multiple times so cache now simply stores `Pkcs12` instance.

sfackler:
I'd like to see some benchmarks indicating that the custom hasher is worth it.

The new behavior does mean the password is ignored after the first decode for a given identity.

blackbeam:
> I'd like to see some benchmarks indicating that the custom hasher is worth it.

[this one](http://aras-p.info/blog/2016/08/02/Hash-Functions-all-the-way-down/)

>The new behavior does mean the password is ignored after the first decode for a given identity.

🤦‍♂️
Well. First of all i tried to simply import pkcs12 without specifying keychain but than i found commit message of 304b00e that states that it'll not work.
Creating another temporary keychain only for password verification looks really ugly.
I ended up with map changed from `der => (Pkcs12, TempDir)` to `(der, hash(pass)) => (Pkcs12, TempDir)` using CommonCrypto bindings.

sfackler:
I'm talking about this use case in particular, not hash functions in general.

blackbeam:
> I'm talking about this use case in particular, not hash functions in general.

Code:
```rust
#[bench]
fn load_pkcs12(bencher: &mut tst::test::Bencher) {
    let buf = include_bytes!("../test/identity.p12");
    bencher.iter(|| Pkcs12::from_der(buf, "mypass").unwrap());
}
```

Default hasher:
```
test test::load_pkcs12 ... bench:       2,249 ns/iter (+/- 333)
```

xxHash:
```
test test::load_pkcs12 ... bench:       1,458 ns/iter (+/- 935)
```

sfackler:
Hmm, did you ever try using a single keychain for all of the identities? It seems like that'd simplify things quite a lot.

blackbeam:
I switched to a single keychain and also have any of password or der cache removed. Also looks like  it is ok to call SecPKCS12Import multiple times on the same der data with the same keychain being specified, however this does not stated clearly in the docs.
I've added `import_same_identity_multiple_times` test end [run it](https://travis-ci.org/blackbeam/rust-native-tls/builds/268885305?utm_source=github_status&utm_medium=notification) on every osx version available on travis.

sfackler:
Awesome, thanks!

PR #65, opened 2018-01-05
Add from_pem() for Certificate

algermissen:
Hi, this is the PR for https://github.com/sfackler/rust-native-tls/issues/64 - in the hopes it is indeed this simple.


sfackler:
It is almost this simple! You'll need to implement the method for the other backends, and then add it to the public type as well:

https://github.com/sfackler/rust-native-tls/blob/master/src/imp/schannel.rs#L93
https://github.com/sfackler/rust-native-tls/blob/master/src/imp/security_framework.rs#L170
https://github.com/sfackler/rust-native-tls/blob/master/src/lib.rs#L192

For schannel you'll call https://docs.rs/schannel/0.1.9/x86_64-pc-windows-msvc/schannel/cert_context/struct.CertContext.html#method.from_pem and for security-framework you'll call http://sfackler.github.io/rust-security-framework/doc/v0.1.10/security_framework/os/macos/import_export/struct.ImportOptions.html, which is a bit more complex.

algermissen:
Ok, will do in my PR

algermissen:
Done - please take a look at the error returned for no cert in security-framework. I am using OSStatus 0 for now because I had no clue which to use. What would be appropriate?

sfackler:
I think `errSecParam` is the best error to return there.

algermissen:
Just made that change. It looks as if the travis build hangs. Do you have the permissions to abort these?

sfackler:
Yeah, it seems like Travis is having networking troubles - I'll adjust some of the CI stuff.

sfackler:
Can you also switch one of the tests over to using from_pem instead of from_der?

algermissen:
Done - created server_pem() from server() test https://github.com/sfackler/rust-native-tls/pull/65/commits/70a7264cfd63211935fb3bebf6df651ba6ed9c2d#diff-c5fc4beef0f03493c2e9a9c876073f0fR92

Let me know if you'd like me to copy more of them - but I think it does not add much value.

sfackler:
Nah, yeah one should be fine.

src/lib.rs
```diff
@@ -195,6 +195,13 @@ impl Certificate {
         let cert = try!(imp::Certificate::from_der(der));
         Ok(Certificate(cert))
     }
+    /// Parses a PEM-formatted X509 certificate.
+    /// If the PEM file contains more than one certificate the last one is used
+    /// and the others are ignored.
```
sfackler: OpenSSL will take the first IIRC. Not sure what Windows will do - I think fail?

algermissen: Your call - I am rather n00b wrt all this - do you want me to change it to take the first?

sfackler: For now let's just leave the behavior unspecified and see if it causes issues. It's pretty broken to depend on the behavior with a multi-cert file anyway.

algermissen:
Sorry, but I have no clue why that import in security-framework version fails. Can you pls take a look?

sfackler:
Ah - that's the iOS build and that security framework API is only available on OSX. We'll probably need to have the iOS version just return an error immediately.

algermissen:
Hi - (not sure if you got notified by the last commit) do you feel the PR is good to go? Would you accept then?

sfackler:
Yep, thanks!

PR #72, opened 2018-01-30
Update lazy_static to 1.0

discosultan:
This PR is similar to https://github.com/sfackler/rust-native-tls/pull/68 but only updates the `lazy_static` dep to avoid breaking the build.

Tests passing on Windows and Ubuntu.


sfackler:
Thanks!

PR #100, opened 2018-06-29
[WIP] Support ALPN on Security Framework

kzys:
This change exposes security-framework's ALPN methods through
native-tls.

security-framework 0.2.1 doesn't have ALPN methods yet.
So the build wouldn't succeed now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -323,6 +323,7 @@ pub struct TlsConnectorBuilder {
     accept_invalid_certs: bool,
     accept_invalid_hostnames: bool,
     use_sni: bool,
+    alpn: Vec<String>,
 }
 
 impl TlsConnectorBuilder {
@@ -405,6 +406,17 @@ impl TlsConnectorBuilder {
         self
     }
 
+    /// Request specific protocols through ALPN (Application-Layer Protocol Negotiation).
+    ///
+    /// Defaults to none
+    pub fn request_protocols(
+        &mut self,
+        protocols: &[&str],
+    ) -> &mut TlsConnectorBuilder {
+        self.alpn = protocols.iter().map(|s| s.to_string()).collect();
+        self
+    }
+
     /// Creates a new `TlsConnector`.
     pub fn build(&self) -> Result<TlsConnector> {
         let connector = imp::TlsConnector::new(self)?;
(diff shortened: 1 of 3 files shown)


kzys:
This branch is still work-in-progress. I'm unsure how to provide the feature on platforms that don't support ALPN (e.g. older OS X versions)

src/lib.rs
```diff
@@ -628,6 +641,11 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         Ok(self.0.tls_server_end_point()?)
     }
 
+    /// Returns the negotiated protocols
+    pub fn negotiated_protocols(&self) -> Result<Vec<String>> {
```
sfackler: ALPN only allows a single protocol to be negotiated, so this should return a `Result<Option<String>>`.

kzys:
Would it make sense to return `Result<(), E>` to tell the caller that the APIs are no-op? Ignoring the call implicitly might confuse developers.

Regarding `dlsym`, how about having that in security-framework? First calling dlsym here may need security-framework to expose some guts such as `SSLContext`. Second people who use security-framework may want to have the dynamic-switch behavior.

sfackler:
Returning an error is a possibility, yeah.

SSLContext is already exposed through the TCFType impl.

kzys:
Well, while SSLContext is available as a type, getting that from secure_transport::SslStream may need some changes in native-tls. Wouldn't it be?

sfackler:
It's directly accessible from here: https://github.com/sfackler/rust-native-tls/blob/master/src/imp/security_framework.rs#L355

sfackler:
We could skip the ClientBuilder and manage the SSLContext directly but there's quite a bit of logic there...

kzys:
Yep. Then again, would it be easier to have dlopen in security-framework? security-framework is currently just a simple one-to-one wrapper for Security.framework. Having dlopen would complicate the crate though.

kzys:
I've opened a new pull request on rust-security-framework, that replaces ALPN-related methods with dlopen-based implementation.

https://github.com/sfackler/rust-security-framework/pull/68

kzys:
Let me close this PR for a moment, since I haven't been working on that for more than a month.

@seanmonstar did similar work for OpenSSL, but we need to make some progress on https://github.com/steffengy/schannel-rs/pull/49 first to have ALPN on rust-native-tls.

PR #109, opened 2018-10-02
Android add certs error: already in hash table

edelangh:
Adding cert on android print error instead of crash

Now when load_android_root_certs failed, it'll print error instead of return error.

ErrorStack([Error { code: 185057381, library: "x509 certificate routines", function: "X509_STORE_add_cert", reason: "cert already in hash table", file: "crypto/x509/x509_lu.c", line: 326 }])

Without, I don't know how can I use hyper_tls on android with certs duplicated

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -103,7 +103,7 @@ fn load_android_root_certs(connector: &mut SslContextBuilder) -> Result<(), Erro
             .filter_map(|b| X509::from_pem(&b).ok());
         for cert in certs {
             if let Err(err) = connector.cert_store_mut().add_cert(cert) {
-                eprintln!("load_android_root_certs error: {:?}", err);
+                debug!("load_android_root_certs error: {:?}", err);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/imp/openssl.rs
```diff
@@ -102,7 +102,9 @@ fn load_android_root_certs(connector: &mut SslContextBuilder) -> Result<(), Erro
             .filter_map(|e| fs::read(e.path()).ok())
             .filter_map(|b| X509::from_pem(&b).ok());
         for cert in certs {
-            connector.cert_store_mut().add_cert(cert)?;
+            if let Err(err) = connector.cert_store_mut().add_cert(cert) {
+                eprintln!("load_android_root_certs error: {:?}", err);
```
sfackler: Can we use `debug!` rather than `eprintln!` here?

Seems reasonable otherwise.

edelangh: Unfortunately rust-native-tls doesn't use crate log

```
error: cannot find macro `debug!` in this scope                                                                       
   --> /tmp/rust-native-tls/src/imp/openssl.rs:106:17                                                                 
    |                                                                                                                 
106 |                 debug!("load_android_root_certs error: {:?}", err);                                             
    |                 ^^^^^                                                                                           
```

Should I add it  ?

sfackler: Yep, we can just add the dependency.

sfackler:
Thanks!

edelangh:
Thank you too!

Issue #49, opened 2017-07-05
ALPN support


alexcrichton:
At the recent all-hands we discussed TLS-in-Rust a bit and one of the nice features to add to this library would be ALPN support, especially in support of HTTP/2. For the backends native-tls has:

* `openssl` crate already has ALPN support
* `security-framework` does not have ALPN support. This may be available in the upcoming high sierra OSX though? And this may be available through private APIs in historical releases of OSX? It may be possible to write a shim that works with all this.
* `schannel` doesn't have ALPN support but it's suspected that it does indeed exist in Windows, we just need to bind it.

hinaria:
hi alex,

here are some notes on using alpn via schannel. hope this helps:

- alpn is indeed available in schannel, but only on windows 8.1 and later. computers running windows 8, windows 7 and windows xp can't use schannel alpn.
- specify alpn protocols by passing a [`SecBuffer { buffer-type = SECBUFFER_APPLICATION_PROTOCOLS }`][1] to [`fn InitializeSecurityContext()`][2].
- query the requested alpn protocols using [`QueryContextAttributes(attribute: SECPKG_ATTR_APPLICATION_PROTOCOL, ...)`][3].

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa379814(v=vs.85).aspx
[2]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa375924(v=vs.85).aspx
[3]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa379340(v=vs.85).aspx

ePirat:
> This may be available in the upcoming high sierra OSX though?
It was promised to be in High Sierra but so far the Security framework does not export the symbols even though the functions are publicly documented. There is a radar filed with Apple about that already.

> And this may be available through private APIs in historical releases of OSX?
Yes there is a private API that could be used since macOS 10.11

sfackler:
ALPN has finally landed in OSX 10.13.4. Unfortunately, it only implements the client side and not the server side so that's all we'd be able to support here.

michaelbeaumont:
Hi, what's the status of this issue? I'm updating a library from 0.1 and I'm stuck now that `backend` was removed (https://github.com/sfackler/rust-native-tls/issues/103). I was downcasting to `openssl` in order to do ALPN negotiation.

est31:
[...]

PR #113, opened 2019-01-03
Add preliminary ALPN support for OpenSSL backend

seanmonstar:


seanmonstar:
Seems CI failures are unhappy about lazy_static 0.2.

sfackler:
I don't really want to support a feature that only works on one backend - the whole point of this crate is that it'll work anywhere!

src/imp/openssl.rs
```diff
@@ -266,6 +266,16 @@ impl TlsConnector {
             connector.cert_store_mut().add_cert((cert.0).0.clone())?;
         }
 
+        let alpn = builder
+            .alpn_protocols
+            .iter()
+            .map(|proto| format!("\\x{:x}{}", proto.len(), proto))
```
sfackler: The sequence that openssl wants is literally the byte length of the protocol followed by it, not the string `\xN`.

seanmonstar:
I assumed as mentioned in #49, support for backends could be added in as they exist.

sfackler:
All of the underlying system APIs support ALPN.

src/imp/openssl.rs
seanmonstar: Ah, my eyes tricked me when reading https://github.com/sfackler/hyper-openssl/blob/master/src/lib.rs#L107.

seanmonstar:
I don't see anything about ALPN in the schannel docs...

sfackler:
https://github.com/steffengy/schannel-rs/pull/49

seanmonstar:
So to clarify, you'd rather not provide the API in `native-tls` until all 3 backends have the features available? I figured we could add support individually, since it's mostly an optimization, and things should still work if not available.

sfackler:
> So to clarify, you'd rather not provide the API in `native-tls` until all 3 backends have the features available?

Yep.

PR #116, opened 2019-01-23
Ignore error in add_cert on openssl

### LINKED ISSUE #115: "cert already in hash table" error could not be ignored
Hi,

Some systems don't have root certificates installed, so that it may be a good idea to provide bunded root CA's as a part of the binary. But, on systems with certificates installed, it may lead to the "cert already in hash table" error. It seems like there is no way to just ignore this error.

glebpom:
Fixes #115

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -263,7 +263,9 @@ impl TlsConnector {
         supported_protocols(builder.min_protocol, builder.max_protocol, &mut connector)?;
 
         for cert in &builder.root_certificates {
-            connector.cert_store_mut().add_cert((cert.0).0.clone())?;
+            if let Err(err) = connector.cert_store_mut().add_cert((cert.0).0.clone()) {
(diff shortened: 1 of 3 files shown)


src/imp/openssl.rs
```diff
@@ -263,7 +263,9 @@ impl TlsConnector {
         supported_protocols(builder.min_protocol, builder.max_protocol, &mut connector)?;
 
         for cert in &builder.root_certificates {
-            connector.cert_store_mut().add_cert((cert.0).0.clone())?;
+            if let Err(err) = connector.cert_store_mut().add_cert((cert.0).0.clone()) {
```
sfackler: I feel like we should do better here and check for the duplicate cert error in particular, but I can do that in a follow-up.

sfackler: Yep!

In practice it's not that big of a problem since I think the only other error that can come out of this method is an allocation failure but it still seems more explicit at least.

PR #117, opened 2019-02-13
TlsStream::certificate_chain, ChainIterator type and Certificate::public_key_info_der

SergejJurecko:
This patch makes certificate pinning possible on top of rust-native-tls. What is required is access to the entire certificate chain and subjectPublicKeyInfo part of a certificate.

This is not ready to be merged yet, as I am waiting for downstream schannel and rust-security-framework patches to be merged. 

I've created this PR now to resolve any blockers in the meantime.

### DIFF
--- a/src/imp/schannel.rs
+++ b/src/imp/schannel.rs
@@ -124,7 +124,7 @@ impl Certificate {
     }
 
     pub fn public_key_der(&self) -> Result<Vec<u8>, Error> {
-        Ok(self.0.subject_public_key_info_der())
+        Ok(self.0.subject_public_key_info_der()?)
     }
 }
 
--- a/src/imp/schannel.rs
+++ b/src/imp/schannel.rs
@@ -1,7 +1,7 @@
 extern crate schannel;
 
 use self::schannel::cert_context::{CertContext, HashAlgorithm};
-use self::schannel::cert_store::{CertAdd, CertStore, Memory, PfxImportOptions};
+use self::schannel::cert_store::{CertAdd, CertStore, Certs, Memory, PfxImportOptions};
 use self::schannel::schannel_cred::{Direction, Protocol, SchannelCred};
 use self::schannel::tls_stream;
 use std::error;
@@ -89,7 +89,8 @@ impl Identity {
                 return Err(io::Error::new(
                     io::ErrorKind::InvalidInput,
                     "No identity found in PKCS #12 archive",
-                ).into());
+                )
+                .into());
             }
         };
 
@@ -115,7 +116,8 @@ impl Certificate {
             Err(_) => Err(io::Error::new(
                 io::ErrorKind::InvalidInput,
                 "PEM representation contains non-UTF-8 bytes",
-            ).into()),
+            )
+            .into()),
         }
     }
 
@@ -153,7 +155,7 @@ where
 
     pub fn handshake(self) -> Result<TlsStream<S>, HandshakeError<S>> {
         match self.0.handshake() {
-            Ok(s) => Ok(TlsStream(s)),
+            Ok(s) => Ok(TlsStream(s, None)),
             Err(e) => Err(e.into()),
         }
     }
@@ -231,7 +233,7 @@ impl TlsConnector {
             builder.verify_callback(|_| Ok(()));
         }
         match builder.connect(cred, stream) {
-            Ok(s) => Ok(TlsStream(s)),
+            Ok(s) => Ok(TlsStream(s, None)),
             Err(e) => Err(e.into()),
         }
     }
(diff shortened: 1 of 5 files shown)


SergejJurecko:
Also I would like to expose hashing algorithms (md5, sha256, sha384 and sha512) from rust-native-tls. Would this be agreeable? It goes a bit further then the stated goal of this project but it's very useful and quite commonly needed for anyone using this upstream.

sfackler:
I have been thinking a bit about adding a `native-crypto` crate as a companion to `native-tls`, yeah. I have less confidence about what a good API for that would be, though. I'm personally more interested in cryptography bits in that context than hashing, since high-quality Rust implementations of sha etc are available.

src/imp/schannel.rs
```diff
@@ -259,13 +265,28 @@ impl TlsAcceptor {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(certs) = self.certs.as_mut() {
+            return certs.next().map(Certificate);
+        }
+        None
+    }
+}
+
+pub struct TlsStream<S>(tls_stream::TlsStream<S>, Option<CertStore>);
```
sfackler: I'd use a struct here with named fields to make things a bit more clear.

src/imp/schannel.rs
```diff
@@ -294,6 +315,24 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         }
     }
 
+    pub fn certificate_chain(&mut self) -> Result<ChainIterator<S>, Error> {
+        if self.1.is_none() {
+            match self.0.peer_certificate() {
+                Ok(cert) => {
+                    self.1 = cert.cert_store();
```
sfackler: Why do we need to cache the cert store?

src/imp/security_framework.rs
```diff
@@ -351,6 +359,24 @@ impl TlsAcceptor {
...
+    _stream: &'a TlsStream<S>,
+}
+impl<'a, S> Iterator for ChainIterator<'a, S> {
+    type Item = Certificate;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(trust) = self.trust.as_ref() {
+            let pos = self.pos;
+            self.pos += 1;
+            return trust.certificate_at_index(pos as _).map(Certificate);
```
sfackler: This needs to stop at the end of the chain, right?

src/imp/security_framework.rs
```diff
@@ -385,6 +411,25 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         Ok(trust.certificate_at_index(0).map(Certificate))
     }
 
+    pub fn certificate_chain(&mut self) -> Result<ChainIterator<S>, Error> {
+        let trust = match self.stream.context().peer_trust2()? {
+            Some(trust) => trust,
+            None => {
+                return Ok(ChainIterator {
```
sfackler: This can be restructured a bit to avoid having to have a separate early return:

```rust
let trust = match self.stream.context().peer_trust2()? {
    Some(trust) => {
        trust.evaluate()?;
        Some(trust)
    }
    None => None
};
...
```

src/lib.rs
```diff
@@ -630,6 +647,11 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         Ok(self.0.peer_certificate()?.map(Certificate))
     }
 
+    /// Returns an iterator over certificate chain, if available.
+    pub fn certificate_chain(&mut self) -> Result<ChainIterator<S>> {
```
sfackler: Should we distinguish the "no chain present" case here by returning something like `Result<Option<ChainIterator>>`?

src/imp/schannel.rs
SergejJurecko: I could not find a better solution. CertStore once you call certs is borrowed. You can't drop it and you can't put it in the iterator as it will be a self referenced struct.

src/imp/security_framework.rs
SergejJurecko: Trust will return None once it is over limit.

src/lib.rs
SergejJurecko: I would prefer to create an empty iterator that just returns None immediately. That is what happens in the downstream crates. Result<Option<ChainIterator>> is a pretty annoying API.

src/imp/schannel.rs
sfackler: It seems like things would be easier to work with if we just returned a `Result<Vec<Certificate>>`. I don't think we're gaining all that much by not buffing the certs up again.

sfackler:
Should we add an intermediate `PublicKey` type instead of directly going from `Certificate` to DER-encoded public key?

SergejJurecko:
Not sure what the API would be for the PublicKey type. schannel-rs looks pretty sparse in this area. Since checks are passing right now, is there anything else to do with regards to this?

SergejJurecko:
Can I do anything more or is this not something you want merged?

edevil:
Having access to the `certificate_chain()` of a TlsStream is important when we're trying to implement something like DANE. Is there something that can be improved here so that it can be merged, or is this not something this project is interested in @sfackler ?

PR #120, opened 2019-03-02
Specify more exact dependencies

### LINKED ISSUE #119: Does not compile on macOS Sierra
The crate has recently stopped compiling on macOS 10.12.6 Sierra. The compiler produces 78 errors, each with the following note:

> note: Perhaps two different versions of crate `core_foundation_sys` are being used?

I believe the issue appeared with today's [security-framework-sys](https://crates.io/crates/security-framework-sys) release.

fenhl:
Fixes #119.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ readme = "README.md"
 vendored = ["openssl/vendored"]
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-security-framework = "0.2.1"
-security-framework-sys = "0.2.1"
+security-framework = "=0.2.1"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -11,8 +11,8 @@ readme = "README.md"
 vendored = ["openssl/vendored"]
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-security-framework = "0.2.1"
-security-framework-sys = "0.2.1"
+security-framework = "=0.2.1"
```
sfackler: = dependencies cause significant issues resolving a dependency graph, and shouldn't be used.

PR #129, opened 2019-07-06
Add a possibility to create system-based Identity

ancwrd1:
This PR adds Identity::from_system function. Currently only works on Windows; other platforms will return Error::Unsupported.

The rationale is that many enterprise applications *require* that the key chains must be taken from secure certificate store without the need to specify the store password.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -115,6 +115,7 @@ fn load_android_root_certs(connector: &mut SslContextBuilder) -> Result<(), Erro
 pub enum Error {
     Normal(ErrorStack),
     Ssl(ssl::Error, X509VerifyResult),
+    Unsupported
 }
 
 impl error::Error for Error {
@@ -122,13 +123,16 @@ impl error::Error for Error {
         match *self {
             Error::Normal(ref e) => error::Error::description(e),
             Error::Ssl(ref e, _) => error::Error::description(e),
+            Error::Unsupported => "Unsupported",
         }
     }
 
+    #[allow(deprecated)]
     fn cause(&self) -> Option<&error::Error> {
         match *self {
             Error::Normal(ref e) => error::Error::cause(e),
             Error::Ssl(ref e, _) => error::Error::cause(e),
+            Error::Unsupported => None
         }
     }
 }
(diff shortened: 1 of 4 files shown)


sfackler:
I don't really want to have APIs that only work on a single platform. Have you considered using schannel directly if you need Windows-specific APIs anyway?

dremon:
> Have you considered using schannel directly if you need Windows-specific APIs anyway?

It requires to reimplement tokio-tls specifically for schannel and then to bring Windows cfg gates into our application which is cross-platform for all 3 major OSes (switch between tokio-tls and tokio-tls-schannel or whatever).

Can be done I guess but the higher-level abstraction is much easier to do because everything is already there (PR is tiny really).

Besides  that I feel like there can be more users asking for this feature, apart from my specific case.

I don't disagree that single-platform APIs are less desired than fully cross-platform stuff but there are some pragmatic practical reasons to consider otherwise.

Well, it's your code anyway, so of course you decide whether to accept or not.

dremon:
Another reason is that the crate is called 'native-tls' giving access to a native API. So would be logical to assume that not all native APIs are the same and shouldn't necessarily be downcasted to a lowest common denominator.
The proposed change can be also placed under `cfg(windows)` gate.

sfackler:
> It requires to reimplement tokio-tls specifically for schannel

That doesn't seem too weird - there are already tokio-openssl and tokio-rustls for example. `tokio-tls` was an unfortunate naming choice, IMO; it should have just been `tokio-native-tls`.

>  then to bring Windows cfg gates into our application which is cross-platform for all 3 major OSes (switch between tokio-tls and tokio-tls-schannel or whatever).

You're already going to have to do that anyways to conditionally call this method on Windows, right?

> downcasted to a lowest common denominator.

That's exactly what this crate is - building the lowest common denominator of things that "just work" for the simple use cases.

dremon:
I am still thinking on a better solution.

Would it make sense perhaps to create a `native-tls-identity` crate which can have various identity implementations and make `native-tls` depend on it instead of providing it's own? I can think of a couple of advantages here like implementation decoupling and a possibility to easily extend it. In this case the new crate might have whatever platform-dependent identities built and the `native-tls` can remain without platform-specific stuff.

sfackler:
But `native-tls` would need to depend on `native-tls-identity`, and that would mean that the versions of all of the platform-specific backends would need to be locked together anyway.

PR #130, opened 2019-07-08
OpenSSL reprobes system CAs

Goirad:
When statically linking to openSSL, we lose access to the system certificates. This PR addresses that issue by having the openSSL TLSConnector probe system CAs on creation.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -111,6 +111,28 @@ fn load_android_root_certs(connector: &mut SslContextBuilder) -> Result<(), Erro
     Ok(())
 }
 
+pub fn with_probed_system_ca_list(connector: &mut SslContextBuilder) -> Result<(), Error> {
+    extern crate openssl_sys;
+    use std::ptr;
+    use std::ffi::CString;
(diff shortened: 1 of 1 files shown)


sfackler:
This is already handled by `init_trust`: https://github.com/sfackler/rust-native-tls/blob/master/src/imp/openssl.rs#L90-L93.

jethrogb:
That seems fine, except it uses a slightly different mechanism to pass the info to OpenSSL. @Goirad can you confirm that it ends up working the same way?

jethrogb:
I orignally wrote the code that uses `SSL_CTX_load_verify_locations` over 2 years ago, and I don't remember why I didn't just use `init_ssl_cert_env_vars`. But I was aware of the function, so there must've been a good reason.

jethrogb:
So I think the problem I was facing before was that I wanted to use the probe results *after* the builder instance was already created. Calling `init_ssl_cert_env_vars` at that time would've had no effect. `native-tls` uses the right order internally, so there's no problem.

Goirad:
Closed - native-tls already does this

PR #134, opened 2019-09-04
Relax bounds for get_ref and get_mut

jonhoo:
It should not be necessary to have `S: Read + Write` to get at references to `S` through `TlsStream<S>`, so this patch removes those bounds. This matters little in practice, since you likely can't construct a `TlsStream` on an `S` that isn't `Read + Write`, but it does mean that downstream crates that wrap `TlsStream` do not have to repeat the `Read + Write` bounds for _their_ `get_` methods.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,6 @@ workflows:
     jobs:
       - linux:
           name: openssl
-          image: 1.31.0-stretch
+          image: 1.32.0-stretch
       - macos:
-          version: 1.31.0
+          version: 1.32.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,6 @@ workflows:
     jobs:
       - linux:
           name: openssl
-          image: 1.30.0-stretch
+          image: 1.31.0-stretch
       - macos:
-          version: 1.30.0
+          version: 1.31.0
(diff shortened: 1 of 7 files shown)


sfackler:
CI should be working if you rebase.

jonhoo:
Rebased! Let's see if it works :)

jonhoo:
Ah, looks like `schannel` adds the same bounds for `get_ref` and `get_mut`:
https://github.com/steffengy/schannel-rs/blob/f593bc896b7e855bd7ccd5bf7e33a94fb3fc84bf/src/tls_stream.rs#L325-L333

jonhoo:
Filed https://github.com/steffengy/schannel-rs/pull/57

jonhoo:
@sfackler https://github.com/steffengy/schannel-rs/pull/57 just landed, so this should now work

sfackler:
Can you update the version constraints on schannel here then?

jonhoo:
@sfackler done!

jonhoo:
The [build failure](https://circleci.com/gh/sfackler/rust-native-tls/253) is:
> error: Edition 2018 is unstable and only available for nightly builds of rustc.
Which I believe is unrelated to my change.

sfackler:
Looks like you'll just need to bump this to 1.31.0: https://github.com/sfackler/rust-native-tls/blob/master/appveyor.yml#L2

jonhoo:
@sfackler Looks like we actually needed 1.32.0 because of changes to `getrandom` :'( Well, CI passes now!

sfackler:
Thanks! Yeah, I tend to not worry too much when we have to bump our MSRV due to a dependency.

PR #137, opened 2019-09-26
Run cargo fmt, fix clippy warnings, update dependencies

Fotosmile:


Cargo.toml
```diff
@@ -13,18 +13,18 @@ vendored = ["openssl/vendored"]
...
-openssl = "0.10.15"
-openssl-sys = "0.9.30"
+log = "0.4.8"
+openssl = "0.10.24"
+openssl-sys = "0.9.49"
 openssl-probe = "0.1"
 
 [dev-dependencies]
-hex = "0.3"
+hex = "0.4"
```
sfackler: All of the bumps here other than hex don't do anything different - the new versions are semver-compatible with the old ones.

PR #142, opened 2019-11-07
Support using mbedtls

jack-fortanix:
Mbedtls is somewhat easier to use in certain situations where no OS facilities are present, such as bare metal or in SGX enclaves. This PR allows building native-tls with mbedtls as an alternative to OpenSSL.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,10 @@ readme = "README.md"
 
 [features]
 vendored = ["openssl/vendored"]
+sgx = ["mbedtls"]
+
+[dependencies]
+mbedtls = { version = "0.5.0", features = ["threading", "pkcs12", "pkcs12_rc2", "std"], optional = true }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 security-framework = "0.3.1"
@@ -20,7 +24,10 @@ tempfile = "3.0"
 [target.'cfg(target_os = "windows")'.dependencies]
 schannel = "0.1.16"
 
-[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))'.dependencies]
+[target.'cfg(feature = "mbedtls")'.dependencies]
+mbedtls = { version = "0.5.0", features = ["threading", "pkcs12", "pkcs12_rc2", "std"], optional = true }
+
+[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios", feature = "mbedtls")))'.dependencies]
 log = "0.4.5"
 openssl = "0.10.25"
 openssl-sys = "0.9.30"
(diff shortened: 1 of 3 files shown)


jack-fortanix:
macOS build failed, seems unrelated to this change

Cargo.toml
```diff
@@ -20,7 +24,10 @@ tempfile = "3.0"
 [target.'cfg(target_os = "windows")'.dependencies]
 schannel = "0.1.16"
 
-[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))'.dependencies]
+[target.'cfg(feature = "mbedtls")'.dependencies]
```
sfackler: You can't cfg on features in the dependency section.

If this is designed for SGX, should it just be filtered to that target_os?

src/imp/mbedtls.rs
```diff
@@ -0,0 +1,546 @@
...
+            accept_invalid_hostnames: builder.accept_invalid_hostnames,
+            use_sni: builder.use_sni
+        })
+    }
+
+    pub fn connect<S>(&self, domain: &str, stream: S) -> Result<TlsStream<S>, HandshakeError<S>>
+    where
+        S: io::Read + io::Write
+    {
+        // If any of the ? fail then memory leaks ...
```
sfackler: That seems bad...

src/imp/mbedtls.rs
```diff
@@ -0,0 +1,546 @@
...
+                Pk::from_private_key(&der, None)
+            }
+            let key = Box::new(keys.pop().unwrap().0.map_err(|_| TlsError::PkInvalidAlg)?);
+
+            Some((cert_chain, key))
+        } else {
+            None
+        };
+
+        unsafe {
```
sfackler: I'm not sure I'm on board with this much unsafe code for just establishing a TLS connection. Why can't the mbedtls crate provide a safe interface?

src/imp/mbedtls.rs
```diff
@@ -0,0 +1,546 @@
...
+unsafe impl<S> Sync for TlsStream<S> {}
+unsafe impl<S> Send for TlsStream<S> {}
+
+impl<S> Drop for TlsStream<S> {
+    fn drop(&mut self) {
+        unsafe {
+            if self.session != ::std::ptr::null_mut() {
+                Box::from_raw(self.session);
+            }
+            Box::from_raw(self.socket);
```
sfackler: Why do these all need to be raw pointers if they're treated as owned data?

src/imp/mbedtls.rs
jethrogb: The current Rust API for MbedTLS doesn't take ownership of anything. So if you run your entire server on a single stack, no unsafe is needed. However, if you want to do something else, you need self-referential structs.

jethrogb:
For reference #211 #261

jethrogb:
For reference #211 #261

PR #144, opened 2019-12-18
Impl/error source

eldesh:
I added `source` method of `std::error::Error` for `Error` and `imp::openssl::Error`.
This PR enable users `downcast_ref` errors like:

```
let tls_error: native_tls::Error = ...
if let Some(ref ssl_error) =tls_error
    .source()
    .and_then(|e| e.downcast_ref::<openssl::ssl::Error>())
{
    ...
}
```

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -133,6 +133,13 @@ impl error::Error for Error {
             Error::Ssl(ref e, _) => error::Error::cause(e),
         }
     }
+
+    fn source(&self) -> Option<&error::Error + 'static> {
+        match *self {
+            Error::Normal(ref e) => Some(e),
+            Error::Ssl(ref e, _) => Some(e),
+        }
+    }
 }
 
 impl fmt::Display for Error {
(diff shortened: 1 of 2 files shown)


src/imp/openssl.rs
```diff
@@ -127,12 +127,19 @@ impl error::Error for Error {
         }
     }
 
-    fn cause(&self) -> Option<&error::Error> {
+    fn cause(&self) -> Option<&dyn error::Error> {
```
sfackler: You can just remove the cause implementations - they'll delegate to source anyway.

sfackler:
You should be aware that we don't consider the underlying TLS implementations to be public API. We can and will e.g. upgrade versions without a breaking release.

eldesh:
It is advisable to provide precise error type from this crate.
But acquiring inner error type using `source` method is reasonable.
Also, the method is required for standard error trait.
A user code depends on an inner error type, breaking can't be avoided.

kornelski:
Done in e3ebe77f

PR #145, opened 2019-12-26
Identity from PEMs

dpmkl:
This adds a convenience method to create an Identity from PEMs, which are provided by LetsEncrypt. 
Furthermore resolves some deprecated warnings.


sfackler:
This will need to be implemented for the other backends.

dpmkl:
Already looking into it

Issue #27, opened 2017-03-07
Switch `TlsAcceptor::builder` input from PKCS#12 to PKCS#8 + cert list


briansmith:
For background, see https://unmitigatedrisk.com/?p=543 and https://www.cs.auckland.ac.nz/~pgut001/cryptlib/faq.html#Q5.

Although it seems simple for rust-native-tls to use PKCS#12 as its standard format for server key/certificate configuration, it is actually quite problematic.

First, PKCS#12 cryptography is stone-aged (3DES, RC2, etc.). This is a huge burden on any crypto library that itself wasn't forged in the stone ages of cryptography. Also, it is common for implementations to use way too few rounds of PBKDF2 (or PKBDF1), making the password-based encryption limited value unless the password itself is a strong random key.

Secondly, the interop is limited due to poor defaults and lack of following recommendations. For example, the AES-CBC encryption that the latest IETF spec recommends isn't supported by most (all?) versions of Windows. As another example, many versions of OpenSSL (and, IIRC, NSS) use RC2 as the default (in some cases only) encryption format. 

Thirdly, it is pretty rare for private keys and certificate chains to be in PKCS#12 format. Usually they are in a collection of unencrypted PEM files. Even in the case where the TLS server does want to encrypt the private key key, usually that encryption is done using a key management system that (in my experience) isn't based on PKCS#12. For example, the ACME protocol (Let's Encrypt's certificate distribution protocol) defaults to PEM-based formats; see https://tools.ietf.org/html/draft-ietf-acme-acme-05#section-6.4.2.

Fourth, PKCS#12 is unnecessarily difficult to implement correctly and safely, and difficult to test and verify.

I propose instead that the constructor for `TlsAcceptor` be changed to accept an unencrypted PKCS#8 private key and a list of certificates (maybe just a list of DER-encoded certificates in `&[u8]` slices). Then support for encrypted key/cert management, including PKCS#12, and also hopefully including better systems than PKCS#12, can be provided by separate crates.

sfackler:
PKCS#12 was not my first choice either, but it wasn't clear to me how to get a `SecIdentity` from separate key and cert files on the macOS side.

shepmaster:
> but it wasn't clear to me how to get a SecIdentity from separate key and cert files on the macOS side.

I'm sure I'm not adding anything that y'all don't already know, but just in case, it looks like `SecIdentityCreateWithCertificate` is useful:

```
    #[test]
    fn identity_pem() {
[...]

PR #147, opened 2020-01-14
Implement PKCS8 certificate support for all three backends.

Goirad:
This solves #27 , although rather than removing support for PKCS12 it just adds support for PKCS8.

cc @jethrogb


src/imp/openssl.rs
```diff
@@ -5,19 +5,19 @@ use self::openssl::error::ErrorStack;
...
 use self::openssl::nid::Nid;
 use self::openssl::pkcs12::Pkcs12;
-use self::openssl::pkey::PKey;
+use self::openssl::pkey::{PKey, Private};
 use self::openssl::ssl::{
     self, MidHandshakeSslStream, SslAcceptor, SslConnector, SslContextBuilder, SslMethod,
     SslVerifyMode,
 };
-use self::openssl::x509::{X509, X509VerifyResult};
+use self::openssl::x509::{X509VerifyResult, X509};
```
jethrogb: Unnecessary change

src/imp/openssl.rs
```diff
@@ -165,7 +165,19 @@ impl Identity {
         Ok(Identity {
             pkey: parsed.pkey,
             cert: parsed.cert,
-            chain: parsed.chain.into_iter().flat_map(|x| x).collect(),
+            chain: parsed.chain.map(|stack| stack.into_iter().collect()),
```
jethrogb: Why this change? Is there a difference between `None` and `Some(vec![])`?

src/lib.rs
```diff
@@ -183,6 +185,13 @@ impl Identity {
         let identity = imp::Identity::from_pkcs12(der, password)?;
         Ok(Identity(identity))
     }
+
+    /// buf is the contents of a file containing a chain of PEM encoded certificates
```
jethrogb: * Don't call the argument `buf`
* Specify the order in which we expect the certificates
* Specify that you expect concatenation

src/lib.rs
```diff
@@ -183,6 +185,13 @@ impl Identity {
...
+    /// key is the contents of a file containing a PEM encoded private key
```
jethrogb: "contents of a file" seems like a weird way to describe this

src/test.rs
```diff
@@ -372,4 +376,43 @@ mod tests {
 
         p!(j.join());
     }
+
+    #[test]
+    fn server_pkcs8() {
```
jethrogb: Test client auth too?

Cargo.toml
```diff
@@ -10,10 +10,13 @@ readme = "README.md"
 [features]
 vendored = ["openssl/vendored"]
 
+[dependencies]
+lazy_static = "1.0"
+rustc-serialize = "0.3"
```
sfackler: Why rustc-serialize?

src/imp/openssl.rs
```diff
@@ -265,7 +282,10 @@ impl TlsConnector {
         if let Some(ref identity) = builder.identity {
             connector.set_certificate(&identity.0.cert)?;
             connector.set_private_key(&identity.0.pkey)?;
-            for cert in identity.0.chain.iter().rev() {
+            for cert in identity.0.chain.iter() {
```
sfackler: I do not believe this change is correct: https://github.com/sfackler/rust-native-tls/commit/05fb5e583be589ab63d9f83d986d095639f8ec44

src/imp/schannel.rs
```diff
@@ -96,6 +97,39 @@ impl Identity {
...
     }
+
+    pub fn from_pkcs8(pem: &[u8], key: &[u8]) -> Result<Identity, Error> {
+        let mut store = Memory::new()?.into_store();
+        let mut cert_iter = crate::pem::PemBlock::new(pem).into_iter();
+        let leaf = cert_iter.next().expect("at least one certificate must be provided to create an identity");
+        let cert = CertContext::from_pem(std::str::from_utf8(leaf).map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "leaf cert contains invalid utf8"))?)?;
+
+        let mut options = AcquireOptions::new();
+        options.container("schannel");
```
sfackler: I seem to remember there being problems on windows if the same container name is used multiple times. Probably worth a test that parses a couple of identities to confirm.

src/pem.rs
```diff
@@ -0,0 +1,220 @@
+#![allow(unused)]
```
sfackler: I'd prefer to lean on the underlying libraries for this logic. It certainly isn't necessary for OpenSSL and Security.framework, and I think schannel probably has APIs that can handle this as well?

src/imp/openssl.rs
Goirad: > When sending a certificate chain, extra chain certificates are sent in order following the end entity certificate.
This is from https://www.openssl.org/docs/manmaster/man3/SSL_CTX_add_extra_chain_cert.html , though looking at https://www.openssl.org/docs/manmaster/man3/PKCS12_parse.html I don't see any documentation about ordering restrictions for `PKCS12_parse`.

src/pem.rs
Goirad: `schannel` and the underlying `wincrypt` APIs don't handle parsing chained PEM encoded certificates, so at least there needs to be functionality for splitting them. Beyond that, the `PrivateKey::from_pkcs8` takes DER, so either I need to convert the PEM to DER in this crate, or add that functionality to `schannel`, probably using https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptstringtobinarya

sfackler: I believe CryptStringToBinary can be made to deal with PEM -> DER parsing.

src/imp/schannel.rs
sfackler: This is the test case that caused issues on my previous attempt: https://github.com/sfackler/rust-native-tls/commit/b0b9164bc2d5fd92c059738979e5e8b59a0df0df

Cargo.toml
jethrogb: I think this is now no longer necessary?

src/pem.rs
```diff
@@ -0,0 +1,113 @@
...
+pub const PEM_CERTIFICATE_REQUEST: &'static PemGuard = pem_guard!("CERTIFICATE REQUEST");
+pub const PEM_ENCRYPTED_PRIVATE_KEY: &'static PemGuard = pem_guard!("ENCRYPTED PRIVATE KEY");
+pub const PEM_PRIVATE_KEY: &'static PemGuard = pem_guard!("PRIVATE KEY");
+pub const PEM_PUBLIC_KEY: &'static PemGuard = pem_guard!("PUBLIC KEY");
+pub const PEM_CMS: &'static PemGuard = pem_guard!("CMS");
+
+const BASE64_PEM_WRAP: usize = 64;
+
+lazy_static!{
+    static ref BASE64_PEM: base64::Config = base64::Config {
```
jethrogb: This is no longer used

src/imp/openssl.rs
jethrogb: There really needs to be a cross-platform test for this behavior.

src/lib.rs
```diff
@@ -27,6 +27,7 @@
 //! * TLS/SSL client communication
 //! * TLS/SSL server communication
 //! * PKCS#12 encoded identities
+//! * PKCS#8  encoded identities
```
jethrogb:
```suggestion
//! * X.509/PKCS#8 encoded identities
```

svartalf:
Hey! I'm quite interested in this PR, as it blocks me from the first release of [my small app](https://github.com/svartalf/unbound-telemetry/issues/1).
Is there anything I can help with to make it happen?

Goirad:
@sfackler it looks like CI is failing due to an issue with the way rustdoc is testing `README.md`. I can't reproduce it locally.

@svartalf I wouldn't mind PR to implement the missing tests described above, the one that sfackler mentioned about reusing the container name or one for checking the certificate ordering.

src/pem.rs
```diff
@@ -0,0 +1,78 @@
+#![allow(unused)]
+
+/// Split data by PEM guard lines
+pub struct PemBlock<'a> {
```
sfackler: Since this is only used by the schannel backend, could you move it into that module?

src/imp/schannel.rs
sfackler: I think this test still needs to be added.

sfackler:
It's probably best to just remove the README test - that feature of rustdoc isn't super robust as the current failure shows.

Keruspe:
@Goirad any chance of splitting the PemBlock part in its own commit on top of master (and maybe give it its own pull request) and rebase the rest on top of that?

It would be useful for #168 too

abonander:
Hey folks, this PR is a bit stale but we'd like to have this functionality for SQLx: https://github.com/launchbadge/sqlx/pull/1166#discussion_r714306831

If you have a certificate and private key, you have to work around this limitation by first bundling them with PKCS#12 and the only crate in Rust that you can do it with (that I'd trust as far as I could throw it) would be the `openssl` crate, but we don't want to require that on platforms where `native-tls` doesn't use it.

Any chance of getting this updated and merged?

sfackler:
PRs welcomed! :D

I think this is (probably?) pretty close to being able to land with a rebase and a couple of bits of feedback resolved.

abonander:
I could probably find some time to take this to the finish line if @Goirad isn't available.

Goirad:
I don't have time to pick this up again, sorry. Please go ahead and pick it up if you find the time.

kazk:
> If you have a certificate and private key, you have to work around this limitation by first bundling them with PKCS#12 and the only crate in Rust that you can do it with (that I'd trust, anyway) would be the `openssl` crate, but we don't want to require that on platforms where `native-tls` doesn't use it.

Yeah, we've been doing this because it's unavoidable. But requiring `openssl` is not enough anymore. macOS Security Framework can't import PKCS#12 with modern encryption algorithms (https://openradar.appspot.com/FB8988319), so this no longer works on macOS if a user chooses the vendored OpenSSL which was recently bumped to 3.0.0. Loading legacy provider is not possible yet https://github.com/sfackler/rust-openssl/pull/1553 (blocked by https://github.com/openssl/openssl/issues/16970).

I've squashed and rebased this PR, then updated the tests to use `test-cert-gen` (https://github.com/sfackler/rust-native-tls/compare/master...kazk:pkcs8), so I can open a new PR later, but I don't know what I'm doing :D

kazk:
Opened #209

PR #157, opened 2020-03-14
support ALPN in client side.

LuoZijun:


sfackler:
This appears to be missing the schannel implementation.

LuoZijun:
@sfackler This https://github.com/steffengy/schannel-rs/pull/68 PR has merged, but not release new version.

src/lib.rs
```diff
@@ -409,6 +411,14 @@ impl TlsConnectorBuilder {
         self
     }
 
+
+    #[cfg(feature = "alpn")]
+    /// Configures the set of protocols used for ALPN.
+    pub fn alpn_protocols(&mut self, protocols: &[&str]) -> &mut TlsConnectorBuilder {
```
cyang1: FYI per RFC 7301, there is no requirement for ALPNs to be UTF-8 decodable.
> Protocols are named by IANA-registered, opaque, non-empty byte strings

See also: [the list of IANA-registered ALPNs](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids) and [RFC 8701](https://tools.ietf.org/html/rfc8701).

LuoZijun: But macOS's implementation of security-framework requires strings

https://docs.rs/security-framework/0.4.1/src/security_framework/secure_transport.rs.html#743-752

cyang1: Yup I agree, I think the macOS API is incorrect here. Given that both Windows and OpenSSL expose the correct APIs, IMO `rust-native-tls` should have the public API also be correct (e.g. `pub fn alpn_protocols(&mut self, protocols: &[&[u8]]) -> &mut TlsConnectorBuilder`), and probably return an `Err` when building the macOS variant if a given ALPN is not UTF-8 decodable.

Just my 2 cents from staring at this for a while, and not as anyone with authority :)

LuoZijun:
```rust
use std::convert::TryFrom;

let mut builder = native_tls::TlsConnector::builder();

// from str (macOS/Windows/Other)
builder.alpn_protocols(&["h2", "dot", "webrtc"]);

// from bytes (Windows/Other)
let data: &[&[u8]] = &[b"h2", b"dot", b"webrtc", b"\xe2\x82\x28"];
if let Ok(alpn) = native_tls::ApplicationProtocols::try_from(data) {
    builder.alpn_protocols(alpn);
}
```

LuoZijun:
@sfackler schannel implementation is done.

Cargo.toml
```diff
@@ -9,22 +9,23 @@ readme = "README.md"
 
 [features]
 vendored = ["openssl/vendored"]
+alpn = [ "security-framework/OSX_10_13", "security-framework-sys/OSX_10_13" ]
```
sfackler: It looks like there's a specific `alpn` feature that can be used instead.

src/alpn.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let protos = &self.protos.inner;
+
+        if self.index >= protos.len() {
+            return None;
+        }
+
+        #[cfg(any(target_os = "macos", target_os = "ios"))]
```
sfackler: All platform specific logic should live in the platform specific submodule.

sfackler:
This needs some tests as well - probably depending explicitly on OpenSSL on the server side and using the native-tls client to talk to it.

LuoZijun:
@sfackler 

Linux ALPN test:
> success.

macOS ALPN test:
> rustls Server     : Custom { kind: UnexpectedEof, error: "tls handshake eof"
> native-tls Client: Failure(Error { code: -4, message: "Function or operation not implemented." })

Windows ALPN test:
> rustls Server     :Custom { kind: InvalidData, error: PeerIncompatibleError("no ciphersuites in common") }
> native-tls Client: Failure(Os { code: -2146893018, kind: Other, message: "The message received was unexpected or badly formatted." })

I guess this has something to do with the version of `tls sdk` that comes with the operating system platform, can you help update the latest system image?

Cargo.toml
```diff
@@ -6,25 +6,39 @@ license = "MIT/Apache-2.0"
...
+default = [ ]
 vendored = ["openssl/vendored"]
+alpn = [ "security-framework/alpn", "security-framework-sys/OSX_10_13" ]
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-security-framework = "0.4.1"
-security-framework-sys = "0.4.1"
-lazy_static = "1.0"
+security-framework = "=0.4.0"
+security-framework-sys = "=0.4.0"
```
goffrie, line 18: Why are these dependencies pinned to an older version?

LuoZijun: problem: https://github.com/kornelski/rust-security-framework/commit/8cdffd81752f68f7a476d768eacadc4e744192c8

`security-framework/Cargo.toml` not add `edition = "2018"`,
and `security-framework/src/lib.rs` not add `extern crate libc;`,
and `security-framework/src/dlsym.rs` has line `use libc;`.

this problem is fixed, but not release.

zonyitoo:
1. openssl passed
2. macos failed. Because of using a deprecated method `keychain`. Requires `security-framework` to release v1.0.0.
3. windows failed. Because of `Custom { kind: InvalidData, error: PeerIncompatibleError("no ciphersuites in common") }`. `rustls` reports this error.

lovasoa:
Hello ! What is the status of this pull request ?

PR #160, opened 2020-03-27
Implement bindings to only use supplied root certificates for validating the server

### LINKED ISSUE #41: Allow distrusting system certificate store
Currently when using CA pinning, you always also trust the system certificate store. This is often not wanted.

cyang1:
Implements a way to distrust the system cert store on all backends.

  - On macOS, this just uses `trust_anchor_certificates_only`.
  - On Windows, this uses a custom `verify_callback` that manually checks that the root cert exists in the user-specified root certs.
  - On OpenSSL, this manually creates a new `X509Store` with the user-specified root certs and sets it with `set_verify_cert_store`. This is the one I'm least sure about -- it is probably more correct to avoid calling `set_default_verify_paths`, but that currently happens on builder construction, and moving where that happens might be a breaking change (as `set_default_verify_paths` reads environment variables, which are set process-wide).

Closes #41.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -10,7 +10,7 @@ use self::openssl::ssl::{
     self, MidHandshakeSslStream, SslAcceptor, SslConnector, SslContextBuilder, SslMethod,
     SslVerifyMode,
 };
-use self::openssl::x509::{X509, X509VerifyResult};
+use self::openssl::x509::{X509, store::X509StoreBuilder, X509VerifyResult};
 use std::error;
 use std::fmt;
 use std::io;
@@ -263,6 +263,10 @@ impl TlsConnector {
             }
         }
         supported_protocols(builder.min_protocol, builder.max_protocol, &mut connector)?;
+
+        if builder.disable_built_in_roots {
+            connector.set_cert_store(X509StoreBuilder::new()?.build());
+        }
 
         for cert in &builder.root_certificates {
             if let Err(err) = connector.cert_store_mut().add_cert((cert.0).0.clone()) {
--- a/src/test.rs
+++ b/src/test.rs
@@ -51,6 +51,51 @@ mod tests {
     }
 
     #[test]
+    fn connect_no_root_certs() {
+        let builder = p!(TlsConnector::builder().disable_built_in_roots(true).build());
+        let s = p!(TcpStream::connect("google.com:443"));
+        assert!(builder.connect("google.com", s).is_err());
+    }
+
+    #[test]
+    fn server_no_root_certs() {
+        let buf = include_bytes!("../test/identity.p12");
+        let identity = p!(Identity::from_pkcs12(buf, "mypass"));
+        let builder = p!(TlsAcceptor::new(identity));
+
+        let listener = p!(TcpListener::bind("0.0.0.0:0"));
+        let port = p!(listener.local_addr()).port();
+
+        let j = thread::spawn(move || {
+            let socket = p!(listener.accept()).0;
+            let mut socket = p!(builder.accept(socket));
+
+            let mut buf = [0; 5];
+            p!(socket.read_exact(&mut buf));
+            assert_eq!(&buf, b"hello");
+
+            p!(socket.write_all(b"world"));
+        });
+
+        let root_ca = include_bytes!("../test/root-ca.der");
+        let root_ca = Certificate::from_der(root_ca).unwrap();
+
+        let socket = p!(TcpStream::connect(("localhost", port)));
+        let builder = p!(TlsConnector::builder()
+            .disable_built_in_roots(true)
+            .add_root_certificate(root_ca)
+            .build());
+        let mut socket = p!(builder.connect("foobar.com", socket));
+
+        p!(socket.write_all(b"hello"));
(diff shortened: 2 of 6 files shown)


src/imp/openssl.rs
```diff
@@ -270,6 +270,16 @@ impl TlsConnector {
...
 
+        if builder.disable_built_in_roots {
+            let mut verify_cert_store = X509StoreBuilder::new()?;
+            for cert in &builder.root_certificates {
+                if let Err(err) = verify_cert_store.add_cert((cert.0).0.clone()) {
+                    debug!("add_cert error: {:?}", err);
+                }
+            }
+            connector.set_verify_cert_store(verify_cert_store.build())?;
+        }
```
goffrie, line 281: This should probably go before the `add_cert`s so it doesn't have to re-add the `root_certificates`?

goffrie: Never mind, apparently `cert_store_mut` accesses a different cert store...

cyang1: Dug into this a bit, I'll summarize the findings here.

Besides the `ctx->cert_store`, OpenSSL has two other optional cert stores, the `cert->verify_store` and the `cert->chain_store`. If the `verify_store` is set, OpenSSL will use it instead of `ctx->cert_store` in [`ssl_verify_cert_chain`](https://github.com/openssl/openssl/blob/e15d369781eb2e97656ebbabef576b8079d86b8c/ssl/ssl_cert.c#L384). If the `chain_store` is set, OpenSSL will use it instead of `ctx->cert_store` in [`ssl_build_cert_chain`](https://github.com/openssl/openssl/blob/e15d369781eb2e97656ebbabef576b8079d86b8c/ssl/ssl_cert.c#L861) and [`ssl_add_cert_chain`](https://github.com/openssl/openssl/blob/beb958ccd87b95f1a86bfe2b879492b62e58d80e/ssl/statem/statem_lib.c#L931).

It does look like we can skip the `connector.cert_store_mut().add_cert((cert.0).0.clone())` above if we use `set_verify_cert_store` and my tests will still pass. Haven't dug far enough into OpenSSL to know why that is.

sfackler: Just to avoid running into any edge cases, let's use SSL_CTX_set_cert_store instead to keep things consistent. I'll add that to rust-openssl and cut a release for you.

goffrie: If we're going to change rust-openssl I think it'd be preferable to opt out of the `set_default_verify_paths` call.

sfackler: That would be a more invasive API change since it currently happens directly in the constructor of the SslConnectorBuilder.

goffrie: True. I suppose it's not a big deal.

sfackler: openssl 0.10.29 has the new method.

src/test.rs
```diff
@@ -50,6 +50,51 @@ mod tests {
         builder.connect("goggle.com", s).unwrap();
     }
 
+    #[test]
+    fn connect_no_root_certs() {
```
rohansharma, line 54: Let's add another assertion that we **can** connect to `google.com` in the default case -- just in case the test suite setup changes to not include those certs in the built in certs.

We should also add a separate test to check that we can actually connect when we supply the right cert.

src/test.rs
```diff
@@ -50,6 +50,51 @@ mod tests {
...
+        let builder = p!(TlsConnector::builder().disable_built_in_roots(true).build());
+        let s = p!(TcpStream::connect("google.com:443"));
+        assert!(builder.connect("google.com", s).is_err());
+    }
+
+    #[test]
+    fn server_no_root_certs() {
```
rohansharma, line 61: Let's add a negative test for this behavior too.

src/test.rs
goffrie: That already exists: https://github.com/sfackler/rust-native-tls/blob/3a29e956c1eadba0364b3071569aa71f9defc1ff/src/test.rs#L23

goffrie: >We should also add a separate test to check that we can actually connect when we supply the right cert.

I'm not sure we want to hardcode the CA cert for google.com since it'll change over time? That's the point of the `server_no_root_certs` test.

rohansharma: Cool, as long as we're confident that we'll never remove `connect_google`, this is fine.

src/test.rs
cyang1: That already exists in `connect_no_root_certs` above, and this test (depending on which negative we're talking about):
https://github.com/sfackler/rust-native-tls/blob/3a29e956c1eadba0364b3071569aa71f9defc1ff/src/test.rs#L231

rohansharma: I synced up with @cyang1 offline. It looks like `server_untrusted` is the negative test we need. This lgtm!

src/imp/openssl.rs
cyang1: Thanks! I'll update shortly.

cyang1:
@sfackler would have a chance to take another look at this?

sfackler:
Thanks!

Issue #161, opened 2020-04-24
TlsAcceptor does not request client certificate?


gxtaillon:
Hello,

I'm trying to use `rust-native-tls` to setup a bare tls connection that would require client to provide a certificate, however I have not found a way of doing this with `TlsAcceptor`. There doesn't seem to be a way to set the verification mode and depth.

The `peer_certificate()` on the call always return Ok(None), and the server never requests the certificate.

Using `openssl s_server -accept *:10000 -cert server.pem -key server.pem -verify 10 -CAfile root-ca.crt`, I can see that when providing a cert on the `TlsConnectorBuilder`, the client sends the certificate.

Looking at the tests, that doesn't seem to be supported.  https://github.com/sfackler/rust-native-tls/blob/75e43daaf9174916611178419805542a9883bae4/src/test.rs#L140

Would you welcome a PR?

sfackler:
Sure

gxtaillon:
Exposing `SslVerifyMode` in some way would clash with the format of boolean settings in `TlsConnectorBuilder`. 

I'd add two bools in `TlsAcceptorBuilder`: `request_client_cert` for SSL_VERIFY_PEER, and `require_client_cert` for SSL_VERIFY_FAIL_IF_NO_PEER_CERT. 

What do you think?

sfackler:
How would this be implemented for schannel and secure-transport?

gxtaillon:
I'm not familiar with schannel and secure-transport. Right now i'd only expose the new settings for openssl.

sfackler:
The entire purpose of this crate is to expose a uniform interface to those three backends. I would like to preserve that.

gxtaillon:
You're not making it easy for yourself

I checked schannel here : https://docs.rs/schannel/0.1.16/schannel/tls_stream/struct.Builder.html
Only way I can find to validate client certs would be to set `verify_callback` and check the `CertValidationResult` for no failed_certificate and for the presence of a certchain.

For security-channel here : https://docs.rs/security-framework/0.4.1/security_framework/secure_transport/struct.SslContext.html
There's `set_client_side_authenticate` with `SslAuthenticate`.

Looking back exposing a type akin to https://docs.rs/security-framework/0.4.1/security_framework/secure_transport/struct.SslAuthenticate.html would be cleaner. There would also be more work needed to allow using the verification callbacks of all three backends, but I don't need that right now, and a type like this would not prevent that from happening.

[...]

PR #162, opened 2020-04-25
Implement client certificate validation

gxtaillon:
### Todo:

Provides a configuration mechanism for client certification verification ie: two-way tls.

- [x] openssl
- [ ] schannel
- [ ] security_framework
- [x] tests

Related to: #161

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -310,6 +310,8 @@ impl TlsAcceptor {
 
         if let Some(client_ca_cert) = &builder.client_cert_verification_ca_cert {
             acceptor.add_client_ca((client_ca_cert.0).0.as_ref())?;
+            // below call is required if the ca is not already trusted
+            acceptor.cert_store_mut().add_cert((client_ca_cert.0).0.to_owned())?;
         }
         let verify_mode = match &builder.client_cert_verification {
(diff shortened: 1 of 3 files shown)


src/imp/openssl.rs
```diff
@@ -307,6 +307,18 @@ impl TlsAcceptor {
         let mut acceptor = SslAcceptor::mozilla_intermediate(SslMethod::tls())?;
         acceptor.set_private_key(&builder.identity.0.pkey)?;
         acceptor.set_certificate(&builder.identity.0.cert)?;
+
+        if let Some(client_ca_cert) = &builder.client_cert_verification_ca_cert {
+            acceptor.add_client_ca((client_ca_cert.0).0.as_ref())?;
+            // below call is required if the ca is not already trusted
+            acceptor.cert_store_mut().add_cert((client_ca_cert.0).0.to_owned())?;
```
gxtaillon, line 314: This line is required to run the tests as-is. Not sure if this should stay in.

cypheratheist:
I just encountered the limitation this is supposed to fix.

I'm writing a decentralized peer-to-peer network, and I need mutual authentication in the communication between the nodes.  Because it's decentralized, there's no CA, so each node does custom per-peer validation.

In short, I need both server and client to exchange certificates -- but the reason I'm writing this is because I also need `danger_accept_invalid_certs()` in both ends.  Are you intending to cover this case as well?

PR #165, opened 2020-05-08
Implement ALPN support

### LINKED ISSUE #49: ALPN support

cyang1:
An alternative to #157 .

Slightly more surgical in a lot of places, doesn't require an extra feature, correct error handling on macOS, and some simple tests.

Closes #49.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
       version:
         type: string
     macos:
-      xcode: "9.0"
+      xcode: "10.0.0"
     environment:
       RUST_BACKTRACE: 1
       RUSTFLAGS: -D warnings
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ readme = "README.md"
 vendored = ["openssl/vendored"]
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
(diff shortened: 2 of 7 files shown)


cyang1:
Ah, ALPN support only exists on macOS 10.13 and above, and the CircleCI config is: https://github.com/sfackler/rust-native-tls/blob/e092843716322479a0434cb3913cea1cde93bea6/.circleci/config.yml#L37
Which according to https://circleci.com/docs/2.0/testing-ios/, (probably) corresponds to macOS 10.12.6.

How do you feel about bumping the CircleCI Xcode version to 9.3.1?

cyang1:
Hmm okay, so with some trial and error, it looks like the symbol was not available in 10.13.3 and is available on 10.13.6 (corresponding to a CircleCI config of Xcode 10.0.0).

PR #167, opened 2020-05-11
Support cipher suite selection

### LINKED ISSUE #4: Support cipher suite selection

cyang1:
The exported API is implemented as an explicit whitelist with a consistent default across platforms based on the allowed ciphersuites in rust-openssl, which is itself based on the list from Python.

The implementation for each backend is roughly what you'd expect.
  - For OpenSSL, it constructs a cartesian product of the selected algorithms, joined together with `+`.
  - For Schannel, it just maps the chosen algorithms to the correct Schannel equivalent.
  - For the Security Framework, it doesn't look like there's any structure behind the ciphersuite constants or more user-friendly APIs for interacting with them, so I've combed through the ciphersuite list and manually compiled lists of cipher suites that use each algorithm.

SHA512 is not included as a possible hash algorithm because it appears to be only available in the Security Framework.

Closes #4.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -11,13 +11,92 @@ use self::openssl::ssl::{
     SslVerifyMode,
 };
 use self::openssl::x509::{X509, store::X509StoreBuilder, X509VerifyResult};
+use std::borrow;
 use std::error;
 use std::fmt;
 use std::io;
 use std::sync::Once;
 
-use {Protocol, TlsAcceptorBuilder, TlsConnectorBuilder};
+use {
+    CipherSuiteSet, Protocol, TlsAcceptorBuilder, TlsBulkEncryptionAlgorithm, TlsConnectorBuilder,
+    TlsHashAlgorithm, TlsKeyExchangeAlgorithm, TlsSignatureAlgorithm,
+};
 use self::openssl::pkey::Private;
+
+const CIPHER_STRING_SUFFIX: &[&str] = &[
+    "!aNULL",
+    "!eNULL",
+    "!IDEA",
+    "!SEED",
+    "!SRP",
+    "!PSK",
+    "@STRENGTH",
+];
+
+fn cartesian_product(
+    xs: impl IntoIterator<Item = Vec<&'static str>>,
+    ys: impl IntoIterator<Item = &'static str> + Clone,
+) -> Vec<Vec<&'static str>> {
+    xs.into_iter()
+        .flat_map(move |x| ys.clone().into_iter().map(move |y| [&x, &[y][..]].concat()))
+        .collect()
+}
+
+fn expand_algorithms(cipher_suites: &CipherSuiteSet) -> String {
+    let mut cipher_suite_strings: Vec<Vec<&'static str>> = vec![];
+
+    cipher_suite_strings.extend(cipher_suites.key_exchange.iter().map(|alg| {
+        vec![match alg {
+            TlsKeyExchangeAlgorithm::Dhe => "DHE",
+            TlsKeyExchangeAlgorithm::Ecdhe => "ECDHE",
+            TlsKeyExchangeAlgorithm::Rsa => "kRSA",
+            TlsKeyExchangeAlgorithm::__NonExhaustive => unreachable!(),
+        }]
+    }));
+
+    cipher_suite_strings = cartesian_product(
+        cipher_suite_strings,
+        cipher_suites.signature.iter().map(|alg| match alg {
+            TlsSignatureAlgorithm::Dss => "aDSS",
+            TlsSignatureAlgorithm::Ecdsa => "aECDSA",
+            TlsSignatureAlgorithm::Rsa => "aRSA",
+            TlsSignatureAlgorithm::__NonExhaustive => unreachable!(),
+        }),
+    );
+    cipher_suite_strings = cartesian_product(
+        cipher_suite_strings,
+        cipher_suites.bulk_encryption.iter().map(|alg| match alg {
+            TlsBulkEncryptionAlgorithm::Aes128 => "AES128",
+            TlsBulkEncryptionAlgorithm::Aes256 => "AES256",
+            TlsBulkEncryptionAlgorithm::Des => "DES",
+            TlsBulkEncryptionAlgorithm::Rc2 => "RC2",
+            TlsBulkEncryptionAlgorithm::Rc4 => "RC4",
+            TlsBulkEncryptionAlgorithm::TripleDes => "3DES",
+            TlsBulkEncryptionAlgorithm::__NonExhaustive => unreachable!(),
+        }),
+    );
+    cipher_suite_strings = cartesian_product(
+        cipher_suite_strings,
+        cipher_suites.hash.iter().map(|alg| match alg {
+            TlsHashAlgorithm::Md5 => "MD5",
+            TlsHashAlgorithm::Sha1 => "SHA1",
+            TlsHashAlgorithm::Sha256 => "SHA256",
+            TlsHashAlgorithm::Sha384 => "SHA384",
+            TlsHashAlgorithm::__NonExhaustive => unreachable!(),
+        }),
+    );
+
+    cipher_suite_strings
+        .into_iter()
+        .map(move |parts| borrow::Cow::Owned(parts.join("+")))
+        .chain(
+            CIPHER_STRING_SUFFIX
+                .iter()
+                .map(|s| borrow::Cow::Borrowed(*s)),
+        )
+        .collect::<Vec<_>>()
+        .join(":")
+}
 
 #[cfg(have_min_max_version)]
 fn supported_protocols(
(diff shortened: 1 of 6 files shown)


src/imp/openssl.rs
```diff
@@ -11,14 +11,93 @@ use self::openssl::ssl::{
...
+            TlsSignatureAlgorithm::Rsa => "aRSA",
+            TlsSignatureAlgorithm::__NonExhaustive => unreachable!(),
+        }),
+    );
+    cipher_suite_strings = cartesian_product(
+        cipher_suite_strings,
+        cipher_suites.bulk_encryption.iter().map(|alg| match alg {
+            TlsBulkEncryptionAlgorithm::Aes128 => "AES128",
+            TlsBulkEncryptionAlgorithm::Aes256 => "AES256",
+            TlsBulkEncryptionAlgorithm::Des => "DES",
```
sfackler, line 71: Do we really need to offer historical curiosities like DES and RC2?

src/imp/openssl.rs
```diff
@@ -11,14 +11,93 @@ use self::openssl::ssl::{
...
+            TlsBulkEncryptionAlgorithm::Rc2 => "RC2",
+            TlsBulkEncryptionAlgorithm::Rc4 => "RC4",
+            TlsBulkEncryptionAlgorithm::TripleDes => "3DES",
```
sfackler, line 74: This should include ChaCha20 IMO.

src/imp/openssl.rs
```diff
@@ -11,14 +11,93 @@ use self::openssl::ssl::{
...
+            TlsKeyExchangeAlgorithm::Ecdhe => "ECDHE",
+            TlsKeyExchangeAlgorithm::Rsa => "kRSA",
+            TlsKeyExchangeAlgorithm::__NonExhaustive => unreachable!(),
+        }]
+    }));
+
+    cipher_suite_strings = cartesian_product(
+        cipher_suite_strings,
+        cipher_suites.signature.iter().map(|alg| match alg {
+            TlsSignatureAlgorithm::Dss => "aDSS",
```
sfackler, line 60: Like below, do we really need DSS?

src/imp/security_framework.rs
```diff
@@ -38,6 +43,485 @@ lazy_static! {
...
+    CipherSuite::TLS_ECDH_ECDSA_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_ECDHE_ECDSA_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_ECDH_RSA_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_RSA_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_DH_DSS_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_DH_RSA_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA,
+    CipherSuite::TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA,
+    // CipherSuite::TLS_PSK_WITH_3DES_EDE_CBC_SHA,
```
sfackler, line 56: Why are some of these commented out?

src/imp/security_framework.rs
```diff
@@ -38,6 +43,485 @@ lazy_static! {
...
+    }
+}
+
+fn expand_algorithms(cipher_suites: &CipherSuiteSet) -> Vec<CipherSuite> {
+    let first = cipher_suites
+        .key_exchange
+        .iter()
+        .flat_map(key_exchange_alg_to_cipher_suites)
+        .copied();
+    let rest: &[HashSet<_>] = &[
```
sfackler, line 499: It might be a bit better to load the key exchange ciphers into a hashset and then iterate through the the signature, bulk_encryption, etc ciphers to progressively remove bad ciphers from the set, and then load the remainder into the returned vec.

src/test.rs
```diff
@@ -417,4 +417,28 @@ mod tests {
 
         p!(j.join());
     }
+
+    #[test]
+    fn badssl_cipher_suites_no_rsa() {
+        let builder = p!(TlsConnector::builder().supported_cipher_suites(
+            // Oddly, on Windows, allowing RSA key exchange, but not RSA signature algorithms still
```
sfackler, line 424: This is the kind of edge case I'm worried about in exposing this level of control over cipher suites.

src/imp/openssl.rs
cyang1: I lean towards not being opinionated about what is available. I can remove them if that's your preference, but I can imagine that someone might have some use for them for connecting to old, insecure servers.

src/imp/openssl.rs
cyang1: That would be nice, but AFAICT Schannel doesn't have any ChaCha20 ciphersuites, and I've just included the GCD of the available algorithms across the different backends.

src/imp/security_framework.rs
cyang1: All the commented out ones are PSK. I can remove them, they're mostly left in as I was checking my work while manually compiling the lists, and for anyone else checking it over to see that it wasn't forgotten, but explicitly excluded.

src/imp/security_framework.rs
cyang1: That's a good point, I can change this.

src/test.rs
cyang1: FWIW, RSA is the only one that's valid in two different parts of the ciphersuite specification, so it's the only one this can happen to. Did some searching and it seems like it might be an expected quirk of Schannel that `RSA_KEYX` implies `RSA_SIGN`, although all the discussions that I found were on tangential topics ([1], [2]).

[1]: https://stackoverflow.com/questions/45211765/cryptoapi-rsa-schannel-provider-error
[2]: https://stackoverflow.com/questions/3568695/ms-csp-difference-between-at-signature-and-rsa-key-sign-and-also-at-keyexchang

src/imp/openssl.rs
sfackler: I think it's fine that schannel doesn't support it yet - many versions of OpenSSL don't either. As long as it's paired with another algorithm that is supported it'll work.

cyang1: Sure, what's the experience you had in mind, though, if the user chooses to only specify ChaCha20 on Windows? As currently written, nothing here returns a `Result`, and from some testing, it looks passing no bulk encryption algorithms to the `supported_algorithms` call in Schannel will result in Windows choosing some (most likely system-wide) defaults. 

On Windows 10 19h1 where I'm trying this out, it looks like it chose to allow AES-128, AES-256, and 3DES, which seems consistent with https://docs.microsoft.com/en-us/windows/win32/secauthn/tls-cipher-suites-in-windows-10-v1903, but probably not ideal as far as the behavior of this API.

sfackler: It would need to fail to handshake, yeah. If that's not easily doable though we can leave it out for now.

src/imp/security_framework.rs
cyang1: Made an attempt at this, although I'm not entirely convinced that it's better that it was before.

rneswold:
Can this be merged? I'm getting flagged by my site security department because I'm advertising 3DES. It would be nice to be able to restrict the ciphers.

k3d3:
I'm interested in this getting merged as well. For the insecure ciphers, perhaps it would be a good idea to call them `DANGER_DES` or something similar, to keep with the trend of prepending danger to things that most people would not want to do?

NikosEfthias:
any updates on this ?

PR #168, opened 2020-05-12
allow parsing several certificates from a single pem

Keruspe:
Is this something we might want to do? it would facilitate importing certificate chains.

Not sure about the security-framework impl which I haven't tested

Not sure how to do that with schannel, suggestions welcome


Keruspe:
The macos failure is unrelated to this issue.
The schannel part could be done by adding a dep on the pem crate which can transform a collection of PEMs in a String into some Vec<Pem>, each element of this vec carrying the corresponding DER. (This could also be used for implement Certificate::from_pem for ios btw)

sfackler:
The functionality seems reasonable to me. Full PEM parsing isn't necessary for schannel, just the ability to identify the individual entries. You can probably pull the logic from another PR: https://github.com/sfackler/rust-native-tls/pull/147/files#diff-f3a4e813fbd2200f75435edfe7dcba3d

Keruspe:
@sfackler Will give this a look, thanks for the pointer!

wrt naming, I didn't try and find some fancy naming and just borrowed the name from openssl. Not sure if we want a better name or if that's well enough

sfackler:
stack_from_pem seems fine for now. I think we'd probably just make this the only interface in the next breaking release and rename to from_pem.

Keruspe:
@sfackler rebased on top on 0.2.9 and added impl for schannel based on PemBlock

sfackler:
It looks like the windows build is red.

Keruspe:
Heh, my bad sorry. Fixed the PemBlock namespacing.

(I do not have access to a windows system though...)

Keruspe:
@sfackler any news on this?

PR #169, opened 2020-05-22
Add force_ssl feature

dubiousjim:
When this feature is selected, attempt to use openssl even on Mac
and Windows systems.

Can be combined with the "vendored" feature, for example if your package already has a dependency like:

    openssl = { version = "0.10", features = "vendored" }

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,23 +8,31 @@ repository = "https://github.com/sfackler/rust-native-tls"
 readme = "README.md"
 
 [features]
+default = ["security-framework", "security-framework-sys", "lazy_static", "libc", "tempfile", "schannel"]
 vendored = ["openssl/vendored"]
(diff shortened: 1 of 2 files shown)


sfackler:
The entire purpose of this library is to use schannel/secure transport on those operating systems. If you want to use OpenSSL, I'd recommend just using the openssl crate.

dubiousjim:
Sure, if I were using this as a direct dependency, it wouldn't make sense. But I'm aiming to use other packages that depend directly or indirectly on this one. It'd be less disruptive to try to get their build processes to pass this flag downstream to `rust-native-tls`, than to try to get them to provide a feature where they instead rely directly on the openssl crate.

sfackler:
Libraries should generally be flexible in which TLS implementation they use. Things hardcoding native-tls is a misfeature IMO, and I don't really want to land features that encourage it.

dubiousjim:
Ok thanks for considering

PR #174, opened 2020-06-11
Update Cargo.toml

burne251:
Any systems running with the latest LibreSSL (3.2.0) will fail to build this package, due to openssl-sys 0.9.55 not supporting it, however this has been fixed in 0.9.58


sfackler:
This is not necessary. Run `cargo update` locally.

PR #182, opened 2020-08-05
fix: Fixed compatibility with openssl-probe

frol:
https://github.com/alexcrichton/openssl-probe/issues/16

The newer version of openssl-probe has a breaking change (0.1.3 has been yanked, but ultimately it may release a new version anyway, so it is better to be prepared)


sfackler:
It seems like we would make that change when upgrading to a new breaking release of the library, not speculatively now.

frol:
Sounds good. Let's keep this PR closed for now, and I will get back once 0.2 is released.

PR #186, opened 2020-09-17
update to latest dependecies version

alex-dukhno:
Updating dependencies to their latest version. The reason is that [grcov](https://github.com/mozilla/grcov) is failing to collect coverage during the build/test due to a bug in https://github.com/servo/core-foundation-rs/pull/357 that is dependency of `security-framework`

@sfackler can you have a look and merge?

Thanks


sfackler:
Thanks!

PR #193, opened 2020-12-11
Allow forcing OpenSSL on Windows and macOS

pimeys:
By disabling default features, and enabling the feature `force-openssl`, allows building the crate linking to OpenSSL on Windows and macOS, and allowing vendoring.

This is a way to use certain TLS installations with native-tls where we do not have a way to change the certificate, and where Apple's Secure Transport does not accept them. In our case, the culprit is Microsoft's SQL Server, Azure SQL and how Secure Transport doesn't accept any of the certificates from Microsoft.

This would be part of solving the issue in https://github.com/prisma/tiberius/issues/65

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,17 +8,27 @@ repository = "https://github.com/sfackler/rust-native-tls"
 readme = "README.md"
 
 [features]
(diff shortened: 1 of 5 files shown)


sfackler:
If you want to force the use of openssl, you should just use openssl.

PR #194, opened 2020-12-11
Alpn Support

### LINKED ISSUE #49: ALPN support

MarnixKuijs:
This is an exact copy of #165, but this one is merge with upstream, so all the credit should go to the original author. I really need ALPN support in native-tls. Sadly the original authors of #165 and #157 don't seem to be actively working on this anymore, so that is why I open this PR so I can get ALPN support in.

Closes #49.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ readme = "README.md"
 vendored = ["openssl/vendored"]
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-security-framework = "0.4.1"
-security-framework-sys = "0.4.1"
+security-framework = { version = "0.4.4", features = ["alpn"] }
+security-framework-sys = "0.4.3"
 lazy_static = "1.0"
 libc = "0.2"
 tempfile = "3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,9 +9,7 @@ readme = "README.md"
 
 [features]
 vendored = ["openssl/vendored"]
-
-# Switch for MacOS to keep support for older versions, Windows targets and openssl target will have alpn regardless
-alpn = ["security-framework/alpn"] 
+alpn = ["security-framework/alpn", "openssl/v102"] 
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 security-framework = "2.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,8 +10,11 @@ readme = "README.md"
 [features]
 vendored = ["openssl/vendored"]
 
+# Switch for MacOS to keep support for older versions, Windows targets and openssl target will have alpn regardless
+alpn = ["security-framework/alpn"] 
+
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-security-framework = {version = "2.0.0", features = ["alpn"]}
+security-framework = "2.0.0"
 security-framework-sys = "2.0.0"
 lazy_static = "1.4.0"
 libc = "0.2"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -328,6 +328,7 @@ pub struct TlsConnectorBuilder {
     accept_invalid_hostnames: bool,
     use_sni: bool,
     disable_built_in_roots: bool,
+    alpn: Vec<Vec<u8>>,
 }
 
 impl TlsConnectorBuilder {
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -376,6 +377,14 @@ impl TlsConnectorBuilder {
         self
     }
 
+    /// Request specific protocols through ALPN (Application-Layer Protocol Negotiation).
+    ///
+    /// Defaults to none
+    pub fn request_alpns(&mut self, protocols: &[&[u8]]) -> &mut TlsConnectorBuilder {
```
sfackler, line 383: This should just take strings since Security framework requires that.

MarnixKuijs:
`request_alpns` now takes a `&[&str]` instead of  `&[&[u8]]` and the builder uses `Vec<String>` instead of `Vec<Vec<u8>>`

Cargo.toml
```diff
@@ -11,7 +11,7 @@ readme = "README.md"
 vendored = ["openssl/vendored"]
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-security-framework = "2.0.0"
+security-framework = {version = "2.0.0", features = ["alpn"]}
```
sfackler, line 14: This will introduce a lower bound of macOS 10.13 - I wonder if it should be flagged as a Cargo feature in case people want to support older versions?

src/lib.rs
```diff
@@ -644,6 +654,14 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         Ok(self.0.tls_server_end_point()?)
     }
 
+    /// Returns the negotiated ALPN protocols
+    pub fn negotiated_alpn(&self) -> Result<Option<String>> {
+        Ok(self
+            .0
+            .negotiated_alpn()?
+            .map(|bytes| String::from_utf8(bytes).unwrap())) //negotiated_alpn always returns valid utf-8
```
sfackler, line 662: Is that really the case? I wouldn't be surprised if e.g. OpenSSL would just return whatever bytes the server sent.

sfackler:
LGTM other than those last few questions!

src/lib.rs
MarnixKuijs: I indeed don't know this for sure not the implementation will use `from_utf8_lossy`

Cargo.toml
MarnixKuijs: Added a feature switch to make sure older MacOS versions are supported. I did enable the feature in the CI since that has an high enough MacOS version.

Cargo.toml
```diff
@@ -10,6 +10,9 @@ readme = "README.md"
 [features]
 vendored = ["openssl/vendored"]
 
+# Switch for MacOS to keep support for older versions, Windows targets and openssl target will have alpn regardless
```
sfackler, line 13: This feature needs to apply to all backends - OpenSSL 1.0.1 and sufficiently old Windows versions won't support it either.

src/lib.rs
```diff
@@ -644,6 +654,15 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         Ok(self.0.tls_server_end_point()?)
     }
 
+    /// Returns the negotiated ALPN protocols
+    pub fn negotiated_alpn(&self) -> Result<Option<String>> {
+        let protocol = self
+            .0
+            .negotiated_alpn()?
+            .map(|bytes| String::from_utf8_lossy(&bytes).into_owned());
```
sfackler, line 662: We shouldn't be transparently changing the values received from the server either. This should probably just return `Vec<u8>`.

MarnixKuijs: Fair enough changed it back.

Cargo.toml
MarnixKuijs: Added alpn feature switch to openssl targets and windows targets.

MarnixKuijs:
Looks like the rustdoc on the readme failed, not due to my changes, but more due to it keeping old artifacts, a clean of the CI should help.

cyang1:
I'm still around :), the original PR just didn't get any feedback unfortunately.

To add some context to the comments here -- IMO `request_alpns` ought to take `&[u8]` and not `&str`, as [RFC 7301](https://tools.ietf.org/html/rfc7301) defines the ALPN identification sequence as any arbitrary sequence of bytes:
```
Identification Sequence: The precise set of octet values that
      identifies the protocol.
```
This is supported by [RFC 8701](https://tools.ietf.org/html/rfc8701), which declares several non-UTF8 ALPN sequences for testing and extensibility purposes.

It's unfortunate that the `Security.framework` API is incorrect on this, but IMO not a reason to propagate this incorrect typing upwards, or make it impossible to specify non-UTF8 ALPNs on other backends.

Also re:
> This will introduce a lower bound of macOS 10.13 - I wonder if it should be flagged as a Cargo feature in case people want to support older versions?

The `OSX_10_13` feature on `rust-security-framework` implies `alpn`, but `alpn` doesn't inherently have any requirements. `rust-security-framework` will dynamically load the symbol if compiled without targeting macOS 10.13 or above.

https://github.com/kornelski/rust-security-framework/blob/c2897426a0d42b8833cb6bf943441a225817e0fd/security-framework/src/secure_transport.rs#L725-L733

Issue #196, opened 2021-01-23
Convert from SslConnector to TlsConnector


massand:
It would be useful to add the ability to construct a [`native_tls::TlsConnector`](https://docs.rs/native-tls/0.2.7/native_tls/struct.TlsConnector.html) from [`openssl:ssl::SslConnector`](https://docs.rs/openssl/0.10.32/openssl/ssl/struct.SslConnector.html) using:
```
From<openssl::ssl::SslConnector> for native_tls::TlsConnector
```

Here is an example scenario - [hyper-proxy](https://docs.rs/hyper-proxy/0.5.0/hyper_proxy/struct.ProxyConnector.html#method.set_tls) provides the ability to set the tunneling `native_tls::TlsConnector`. However, when employing custom OpenSSL engines to offload crypto operations during TLS, the openssl crate is the only convenient path to achieve such an implementation.  

It would be cleaner to add support for the `openssl::ssl::SslConnector -> native_tls::TlsConnector` conversion, which would allow  hyper-proxy to continue using native_tls::TlsConnector (and tokio-native-tls stream) for TLS handshake, instead of adding a new [hyper-proxy feature](https://github.com/tafia/hyper-proxy#features) for `openssl::ssl::SslConnector` (and [`tokio_openssl::SslStream`](https://docs.rs/tokio-openssl/0.6.1/tokio_openssl/struct.SslStream.html)) support.

sfackler:
This crate does not publicly expose the backend libraries so that they can be upgraded without a breaking change. This crate is specifically designed to be a "least common denominator", not an general interface on top of the backend libraries.

ryankurte:
Would you be open to adding this behind a feature guard? I understand the intent is to isolate underlying TLS library but, `native-tls` appears used super commonly as a transitive dependency often exposed in the API of other crates. Because of the least common denominator approach if you need any additional configuration you cannot use -any- of them (at least, without refactoring them to not use `native-tls`).

sfackler:
Adding a Cargo feature does not change the fact that exposing the underlying libraries would make it a breaking change to upgrade them. This crate is not intended to be the single canonical TLS library that everything uses.

PR #197, opened 2021-02-01
Support converting openssl::ssl::SslConnection to native_tls::TlsConnection

ryankurte:
I'm not sure if y'all are interested in this as an addition, but, it's got me out of a dependency cycle with openssl/native_ssl.

This lets one to create a `native_tls::TlsConnection` directly from an existing `openssl::tls::TlsConnection` to allow low-level configuration of options not exposed by native-tls when using openssl. This is specifically useful where intermediate crates expect a native-tls context that cannot otherwise be created.


sfackler:
#196

PR #202, opened 2021-07-16
Enable partial certificate chain validation

nicklan:
This adds code to the Openssl implementation to set the `X509_V_FLAG_PARTIAL_CHAIN` flag (see [here](https://www.openssl.org/docs/man1.1.0/man3/X509_VERIFY_PARAM_set_flags.html)).

The way I've written it, it defaults to true. This is the behavior of curl ([see here](https://daniel.haxx.se/blog/2020/01/08/)), and also seems to match the default go http client. The curl page also claims this is the more common behavior.

If you'd possible accept this I can see what happens on mac/windows, but I couldn't find anything related to it in the schannel or security-framework docs. I'm not sure what their default behavior is, and testing on those platforms is a bit tricky, but I can figure it out if needed.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -11,6 +11,7 @@ use self::openssl::ssl::{
     SslVerifyMode,
 };
 use self::openssl::x509::{store::X509StoreBuilder, X509VerifyResult, X509};
+use self::openssl::x509::verify::X509VerifyFlags;
 use std::error;
 use std::fmt;
 use std::io;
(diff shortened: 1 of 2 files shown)


sfackler:
Configuration options need to apply to every backend.

nicklan:
i figured that would be an issue. so what about really useful features like this one that may not? just give up?

I'm guessing `#cfg`ing the higher level methods on the builder aren't an option?

nicklan:
btw, [this page](https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html) suggests to me that at least schannel defaults to this behavior, since curl provides an option to turn it OFF, which is only available to the openssl backend.

if that was the case, would you accept a pr to bring the openssl implementation in line with the schannel one?

PR #204, opened 2021-08-08
Fix off-by-one error in schannel's max_protocol_version

blyxxyz:
`max_protocol_version` set a maximum of one version earlier than it was supposed to.

There are two tests that exercise this. I believe they didn't catch it because:
- `server_tls11_only` sets the same `min_protocol_version` and `max_protocol_version`, which causes `schannel` to be configured with no versions at all, which flips it into its default setting (?) which the test also accepts.
- `server_no_shared_protocol` was supposed to fail and still fails if you make it even more restrictive.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,5 @@ jobs:
         with:
           path: target
           key: target-${{ runner.os }}-${{ steps.rust-version.outputs.version }}-${{ hashFiles('Cargo.lock') }}
-      - run: cargo test
-      - run: cargo test --features alpn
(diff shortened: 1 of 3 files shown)


sfackler:
The ubuntu failure seems unrelated to this change. I wonder if they disabled TLS 1.1 in their build? Might be easiest to just enable the `vendored` feature in that build setup.

blyxxyz:
Sounds good. The feature doesn't affect the other platforms, so I've just enabled it unconditionally.

sfackler:
Thanks!

Issue #110, opened 2018-10-21
TlsConnector adds certificate chain from PKCS12 identity in reverse order


dremon:
Using native-tls 0.2.1 on Linux (openssl).

When building TlsConnector with pkcs12 identity which contains 3 certificates: [leaf, intermediate, root] the intermediate and root certificates are sent to the server peer in reverse order [leaf, root, intermediate] resulting in handshake errors (in particular the javax.net.ssl.SSLException: subject/issuer name chaining check failed).

sfackler:
What order are the certificates stored in the PKCS12 file?

dremon:
> #111 should fix this - could you try it out to confirm?

I think it's only implemented for TlsAcceptor in this PR, should be also changed for TlsConnector. When I change it manually it works, yes.

sfackler:
Great! I'll cut a release with the fix this morning.

PR #209, opened 2021-11-07
Add PKCS8 Support

### LINKED ISSUE #27: Switch `TlsAcceptor::builder` input from PKCS#12 to PKCS#8 + cert list

kazk:
Squashed and rebased #147, then updated the tests to use `test-cert-gen` like the other tests.

Closes #27

---

- [x] Reject other formats in `openssl` and `security_framework`. Done by simply checking the start of the key.
- [x] Update `from_pkcs8` tests to use PKCS#8 keys. `test_cert_gen` doesn't support it, so generated RSA keys are converted using `openssl pkcs8 -topk8 -nocrypt`.
- [x] Fix Windows failing `two_servers` test. Fixed by using unique container names.


src/imp/openssl.rs
```diff
@@ -161,6 +160,19 @@ impl Identity {
             chain: parsed.chain.into_iter().flatten().collect(),
         })
     }
+
+    pub fn from_pkcs8(buf: &[u8], key: &[u8]) -> Result<Identity, Error> {
+        let pkey = PKey::private_key_from_pem(key)?;
+        let mut cert_chain = X509::stack_from_pem(buf)?.into_iter();
+        let cert = cert_chain.next();
+        let chain = cert_chain.collect();
```
sfackler, line 168: Isn't this collecting a vec directly back into a vec?

src/imp/openssl.rs
```diff
@@ -258,7 +270,10 @@ impl TlsConnector {
         if let Some(ref identity) = builder.identity {
             connector.set_certificate(&identity.0.cert)?;
             connector.set_private_key(&identity.0.pkey)?;
-            for cert in identity.0.chain.iter().rev() {
+            for cert in identity.0.chain.iter() {
```
sfackler, line 273: Why is this order being reversed?

src/imp/schannel.rs
```diff
@@ -93,6 +94,38 @@ impl Identity {
 
         Ok(Identity { cert: identity })
     }
+
+    pub fn from_pkcs8(pem: &[u8], key: &[u8]) -> Result<Identity, Error> {
+        let mut store = Memory::new()?.into_store();
+        let mut cert_iter = pem::PemBlock::new(pem).into_iter();
+        let leaf = cert_iter.next().expect("at least one certificate must be provided to create an identity");
```
sfackler, line 101: This shouldn't panic.

src/imp/openssl.rs
```diff
@@ -161,6 +160,19 @@ impl Identity {
...
+
+    pub fn from_pkcs8(buf: &[u8], key: &[u8]) -> Result<Identity, Error> {
+        let pkey = PKey::private_key_from_pem(key)?;
+        let mut cert_chain = X509::stack_from_pem(buf)?.into_iter();
+        let cert = cert_chain.next();
+        let chain = cert_chain.collect();
+        Ok(Identity {
+            pkey,
+            // an identity must have at least one certificate, the leaf cert
+            cert: cert.expect("at least one certificate must be provided to create an identity"),
```
sfackler, line 172: This shouldn't panic.

src/imp/schannel.rs
```diff
@@ -93,6 +94,38 @@ impl Identity {
...
+            .container("schannel")
+            .type_(type_)
+            .keep_open(true)
+            .key_spec(KeySpec::key_exchange())
+            .set()?;
+        let mut context = store.add_cert(&cert, CertAdd::Always)?;
+
+        for int_cert in cert_iter {
+            let certificate = Certificate::from_pem(int_cert)?;
+            context = store.add_cert(&certificate.0, schannel::cert_store::CertAdd::Always)?;
```
sfackler, line 125: `CertAdd::Always`.

src/imp/security_framework.rs
```diff
@@ -82,6 +84,33 @@ pub struct Identity {
 }
 
 impl Identity {
+    pub fn from_pkcs8(pem: &[u8], key: &[u8]) -> Result<Identity, Error> {
+        let dir = TempDir::new().unwrap();
```
sfackler, line 88: This shouldn't panic.

src/imp/security_framework.rs
```diff
@@ -82,6 +84,33 @@ pub struct Identity {
...
+            .keychain(&keychain)
+            .import(&key)?;
+
+        ImportOptions::new()
+            .filename("chain.pem")
+            .items(&mut items)
+            .keychain(&keychain)
+            .import(&pem)?;
+
+        let ident = SecIdentity::with_certificate(&[keychain], &items.certificates[0])?;
```
sfackler, line 107: This shouldn't panic if certificates is empty.

src/imp/openssl.rs
kazk: I don't know the details. Previous discussion: https://github.com/sfackler/rust-native-tls/pull/147#discussion_r367203922

src/imp/openssl.rs
kazk: It's collecting the rest of the chain.

src/imp/openssl.rs
kazk: In https://github.com/sfackler/rust-native-tls/commit/05fb5e583be589ab63d9f83d986d095639f8ec44 you wrote:

> Reverse chain loading for openssl
> The stack is the reverse of what you might expect due to the way
PKCS12_parse is implemented, so we need to load it backwards.
> Closes #110

If `PKCS12_parse` is the cause, maybe it's better to reverse the chain in `from_pkcs12`?

https://github.com/sfackler/rust-native-tls/blob/82ae9a35f92b3e86f44ee5c7245aa991eb8e63d2/src/imp/openssl.rs#L161

kazk: Added [`7b7afc2` (#209)](https://github.com/sfackler/rust-native-tls/pull/209/commits/7b7afc2b7b290f0b075a066cd61ab1a4815379dd)

xinyufort:
I noticed this PR has been sitting for a few weeks now. This is also something that I'm hoping can land soon, since. Is there anything I can do to help get these changes merged into master?

(And thank you @kazk @Goirad for all the work you've put in so far!)

kazk:
@sfackler Please approve running workflows.

src/imp/security_framework.rs
```diff
@@ -428,8 +428,11 @@ impl<S: io::Read + io::Write> TlsStream<S> {
             Some(trust) => trust,
             None => return Ok(None),
         };
-        trust.evaluate()?;
+        trust.evaluate_with_error()?;
```
sfackler, line 431: This won't build, as the return type is not base::Error.

sfackler:
Here's a patch to fix the macOS build:

```diff
diff --git a/src/imp/security_framework.rs b/src/imp/security_framework.rs
index ca86cd5..f133903 100644
--- a/src/imp/security_framework.rs
+++ b/src/imp/security_framework.rs
@@ -423,6 +423,7 @@ impl<S: io::Read + io::Write> TlsStream<S> {
         Ok(self.stream.context().buffered_read_size()?)
     }
 
+    #[allow(deprecated)]
     pub fn peer_certificate(&self) -> Result<Option<Certificate>, Error> {
         let trust = match self.stream.context().peer_trust2()? {
             Some(trust) => trust,
diff --git a/src/test.rs b/src/test.rs
index 897bb85..5fe8151 100644
--- a/src/test.rs
+++ b/src/test.rs
@@ -186,7 +186,10 @@ fn peer_certificate() {
     let socket = p!(builder.connect("localhost", socket));
 
     let cert = socket.peer_certificate().unwrap().unwrap();
-    assert_eq!(cert.to_der().unwrap(), keys.client.ca.get_der());
+    assert_eq!(
+        cert.to_der().unwrap(),
+        keys.server.cert_and_key.cert.get_der()
+    );
 
     p!(j.join());
 }
```

sfackler:
I will have to fix the OpenSSL issue tomorrow.

sfackler:
Linux tests should be fixed if you merge in master.

kazk:
@sfackler Thanks, rebased on the latest master. I need you to approve running workflows again :|

kazk:
`schannel` doesn't support importing PEM encoded RSA key at the moment. https://github.com/steffengy/schannel-rs/pull/64#issuecomment-576995977

I don't have Windows machine to test, but looks like it can be decoded with

```rust
let res = wincrypt::CryptDecodeObjectEx(wincrypt::X509_ASN_ENCODING |
                                        wincrypt::PKCS_7_ASN_ENCODING,
                                        wincrypt::PKCS_RSA_PRIVATE_KEY, // <-
                                        der.as_ptr(),
                                        der.len() as winapi::DWORD,
                                        wincrypt::CRYPT_DECODE_ALLOC_FLAG,
                                        ptr::null_mut(),
                                        &mut buf as *mut _ as winapi::LPVOID,
                                        &mut len);
```

jethrogb:
I'm not sure why you say PEM is not supported and then link to a PR implementing exactly that? Are you trying a PKCS#1 or a PKCS#8 key?

kazk:
I linked to the comment saying:

> Turns out it rejects PKCS1 keys. I've added a test to that effect.

They added explicit check:
```rust
    /// Imports a PEM-encoded PKCS8 private key.
    /// This functions decodes PEM blocks with or without "-----BEGIN PRIVATE KEY-----"
    /// and "-----END PRIVATE KEY-----" headers, but if PEM guards are present they must be exactly
    /// these.
```

Tests are failing on Windows here because we're testing with PEMs having `BEGIN RSA PRIVATE KEY` (PKCS1) headers.

As far as I can tell, "Turns out it rejects PKCS1 keys." is because it failed to decode DER encoded key. The snippet I posted above should be able to decode RSA key, and `schannel` should be able to support both by using the headers to detect the kind.

jethrogb:
Where is the key coming from? I don't see it in the PR.

Anyway, the feature is to add PKCS8 support, the test should use a PKCS8 key.

kazk:
Keys are generated by `test_cert_gen` crate.

kazk:
> Anyway, the feature is to add PKCS8 support, the test should use a PKCS8 key.

The function should probably be renamed. `schannel` is the only one that ignores the headers in PEM (should be possible to fix this). I think most users want a way to create an `Identity` from a certificate chain and a private key for client authentication, and not PKCS8 support specifically. I don't think it's a good idea to limit to PKCS8 key because users will have to parse the PEM and select a specific function to call (unless `native_tls` adds a wrapper function later), and `native_tls` will have to add separate functions wrapping each backend.

`rustls` supports creating config with a certificate chain, and a matching private key. They can add support for more key types without breaking changes. I think `native-tls` should provide something similar. The key types supported by each backend may vary, but that's already the case with PKCS12.

jethrogb:
`native-tls` is about providing standard, cross-platform primitives. I strongly believe we should standardize on PKCS8 for the private key format as it supports all key types. I also think blindly accepting different formats in the same function adds to the general confusion about key formats. Adding PKCS1 support separately may be done but I don't really see the appeal. It's pretty easy to convert PKCS1 keys to PKCS8.

> users will have to parse the PEM and select a specific function to call

I don't understand what you mean by this.

kazk:
Just to be clear, only accepting PKCS#8 is fine if that's what is best for `native_tls`. I'm not against it. I just thought it's more ergonomic to have a function to create an `Identity` from a certificate chain and a PEM-encoded private key, and thought it's easier for `native-tls` because it can let each backend handle the PEM. This PR (#147) currently works that way, and `from_pkcs8` is misleading.

If we're going strict, we'll need to reject keys in `openssl` and `security_framework`.

> I don't understand what you mean by this.

If `native_tls` provided separate functions for each key `from_pkcs8`/`from_rsa`/`from_ecdsa` etc., each user (library) will have to parse the PEM to find which one to call. It's pretty common to not have control over the private key format.

> Adding PKCS1 support separately may be done but I don't really see the appeal. It's pretty easy to convert PKCS1 keys to PKCS8.

If `native-tls` can provide a cross-platform way to convert between formats, that'll be great.

sfackler:
I would be more than happy to also support PKCS1 (if for no other reason than to avoid all of the support issues from people with the wrong encoding :D), but whatever is implemented does definitely need to be consistent across the backends, either supporting or not supporting a format.

xinyufort:
Oh wow... thank you all for the flurry of comments and commits!

While I'm mainly looking for PKCS 8 support (since in my case, I can require that the private keys are in PKCS 8 form), I agree with @kazk and @sfackler that if we are to limit the key format to PKCS 8, we need to do so consistently across all backends, and if we are to support something, all three backends should support it. (I'd also add that if we want to make a generic function that accepts keys of different formats (PKCS 1, PKCS 8, SEC 1, etc.), it shouldn't be named `from_pkcs8`, obviously :slightly_smiling_face:)

And again, do let me know if there's anything I can do to help!

kazk:
I'll change to only support PKCS#8 format by rejecting anything else. I don't have a macOS/Windows machine to work on adding and testing other formats. Also, I found a way to convert PKCS#1 and SEC1 to PKCS#8 in pure Rust that might be good enough for my use case (started [topk8]).

I tried using [topk8] to [test Windows after converting PKCS#1 to PKCS#8](https://github.com/kazk/rust-native-tls/runs/4489650180?check_suite_focus=true), and it seems to work. But `two_servers` test panicked with:
- `"unexpected EOF during handshake"` on accept
- `"The Local Security Authority cannot be contacted"` on connect

Not sure what's wrong because `server_pkcs8` test passes.

---

TODO

- [ ] Reject other formats in `openssl` and `security_framework`
- [ ] Update `from_pkcs8` tests to use PKCS#8 keys. `test_cert_gen` doesn't support it :(
- [ ] Fix Windows failing `two_servers` test

[topk8]: https://github.com/kazk/topk8

Cargo.toml
```diff
@@ -34,6 +34,11 @@ openssl-probe = "0.1"
 [target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))'.build_dependencies]
 openssl-src = { version = "300.0.3", optional = true }
 
+[dependencies]
+pem-rfc7468 = { version = "0.2.4", features = ["std"] }
+
 [dev-dependencies]
+pem = "1.0"
+rsa = { version = "0.5.0", features = ["alloc", "pem", "std"] }
```
kazk, line 42: Used to convert RSA keys from `test-cert-gen` to PKCS#8 in tests for `from_pkcs8`.

kazk:
@sfackler I need you to approve running workflows again. I [tested on my fork](https://github.com/kazk/rust-native-tls/actions/runs/1566379818), and all tests should pass.

src/imp/schannel.rs
```diff
@@ -93,6 +95,61 @@ impl Identity {
...
+        }
+        Ok(Identity { cert: context })
+    }
+}
+
+// The name of the container must be unique to have multiple active keys.
+fn gen_container_name() -> String {
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    static COUNTER: AtomicUsize = AtomicUsize::new(0);
+    format!("native-tls-{}", COUNTER.fetch_add(1, Ordering::Relaxed))
```
kazk, line 152: `two_servers` test fails without this.

Cargo.toml
jethrogb: Can you just use a test key that's already in the right format?

kazk: Rewrote without new dependencies.

jethrogb:
I have no experience with MacOS programming, and I see the P12 code also does this, but is there really no way to import key material without involving the filesystem?

sfackler:
Yeah, unless things have changed since last time I looked into this, you can only get a SecIdentify out of a keychain, and keychains have to live on the filesystem.

src/imp/security_framework.rs
```diff
@@ -82,6 +84,41 @@ pub struct Identity {
...
 
 impl Identity {
+    pub fn from_pkcs8(pem: &[u8], key: &[u8]) -> Result<Identity, Error> {
+        if !key.starts_with(b"-----BEGIN PRIVATE KEY-----") {
+            return Err(Error(base::Error::from(errSecParam)));
+        }
+
+        let dir = TempDir::new().map_err(|_| Error(base::Error::from(errSecIO)))?;
+        let keychain = keychain::CreateOptions::new()
+            .password("password")
```
sfackler, line 94: It would probably be better to generate a random password for each of these keychains since they will be visible on disk for a small period of time.

kazk: Done. By the way, would you accept a separate PR after this to use the 2018 edition?

sfackler: Sure

Cargo.toml
```diff
@@ -21,6 +21,7 @@ security-framework-sys = "2.0.0"
 lazy_static = "1.4.0"
 libc = "0.2"
 tempfile = "3.1.0"
+rand = "0.8.4"
```
sfackler, line 24: Can you use https://docs.rs/security-framework/latest/security_framework/random/struct.SecRandom.html instead to avoid the extra dependency? You should be able to just hex encode the randomness or whatever to make a string.

kazk: `tempfile` depends on `rand` already, so this is not a new dependency. But we can use `SecRandom` instead.

jethrogb:
What needs to be done before merge?

PR #220, opened 2022-03-31
Use 2018 edition

kazk:
https://github.com/sfackler/rust-native-tls/pull/209#discussion_r767399913


kornelski:
Done in https://github.com/rust-native-tls/rust-native-tls/pull/341

PR #223, opened 2022-04-20
Borrow the bytes array for Rust < 1.53

NinoScript:
Addresses issue https://github.com/sfackler/rust-native-tls/issues/222 to keep MSRV lower than 1.53


sfackler:
Can you also update the version we test against in CI? https://github.com/sfackler/rust-native-tls/blob/master/.github/workflows/ci.yml#L39

NinoScript:
CI failed with the lower MSRV.
Other projects need changes too to keep the MSRV low.
I decided that I'd rather push for updating Rust than force open-source projects to use older syntax.

PR #229, opened 2022-06-11
Add ALPN support to TlsAcceptor

matszpk:
Add ALPN to TlsAcceptor - in server side.


sfackler:
This needs to be implemented for all backends. In particular, at least last time I checked Security.framework didn't support server-side ALPN.

JohnGu9:
I think it is ok just let `alpn` be a optional function for difference platform. If Security.framework didn't support server-side ALPN, just *ignore the alpn settings* or *create a runtime error when set server side alpn* on ios and macOS.

PR #234, opened 2022-09-16
Allows the user to implement the logic for identity object creation

msabansal:
Similar to whats mentioned in https://github.com/sfackler/rust-native-tls/issues/232 we have a scenario where we want to use certificates which are non-exportable.
Coming up with a clean cross platform design is challenging. This is our attempt to keep things simple and delegate the complexity to the user

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -184,6 +184,10 @@ impl Identity {
         let chain = cert_chain.collect();
         Ok(Identity { pkey, cert, chain })
     }
+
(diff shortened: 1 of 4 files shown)


sfackler:
I do not want to expose the underlying implementing crates to the user.

msabansal:
While I understand the reasoning for it designing a cross platform feature for something that an abstraction over the underlay feels non trivial.

If it is acceptable we can hide this behind a feature flag

sfackler:
It is absolutely nontrivial, but that is what this library is.

PR #159, opened 2020-03-25
Tls13

olback:
I probably missed something somewhere but it's a start.

Issue #140, opened 2019-10-09
Support TLS 1.3?


driftluo:
I observed that openssl already supports TLS 1.3, security-framework does not support TLS 1.3 and native-tls does not currently support it：
https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/mod.rs#L587
https://github.com/kornelski/rust-security-framework/blob/master/security-framework/src/secure_transport.rs#L383-L419

https://github.com/sfackler/rust-native-tls/blob/7d34900136243a19481a93d86351ba87e6d5a9bf/src/lib.rs#L311-L327

When is this work scheduled to start and finish? has a plan?

sfackler:
It may be challenging to make sure a proper error is produced if the user sets the min version to 1.3 when using the schannel or security-framework backends that don't even know about that version.

driftluo:
As you might expect, I actually want to lock on tls 1.3, don't support for lower versions.

emmm... it looks like a good idea, but just openssl/libressl/boringssl support it :(

mehcode:
TLS 1.3 is supported on macOS 10.14+ and iOS 12.2+ and since about mid-2019 on Windows 10.

https://github.com/retep998/winapi-rs/pull/859

https://github.com/kornelski/rust-security-framework/pull/93

jedisct1:
Now that macOS and Windows have had support for TLS 1.3 for a while, is it something `native_tls` could allow?

More and more servers are now only accepting TLS 1.3.

guytbk:
Hey, is there any update on these? Is there intention to support tls1.3?

roguh:
Also looking for any updates.

briansmith:
As of today, Microsoft [says](https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-):
> TLS 1.3 is supported starting in Windows 11 and Windows Server 2022. Enabling TLS 1.3 on earlier versions of Windows is not a safe system configuration.

There are registry settings to enable TLS 1.3 in Windows 10 but apparently Microsoft doesn't consider using them to be safe; elsewhere, they say that the Windows 10 TLS 1.3 is for testing only.

So to support Windows 10 users safely one either has to be willing to do TLS 1.2 or use a different stack.

Personally, I think it is much more important to get it working for Android and iOS because generally we have more binary size pressure for those targets. Accordingly, I wonder if we could break this into parts by OS.

[...]

PR #235, opened 2022-09-20
Enable TLS 1.3

RyanMeulenkamp:
Retry of #159

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -55,8 +55,7 @@ fn supported_protocols(
         | SslOptions::NO_SSLV3
         | SslOptions::NO_TLSV1
         | SslOptions::NO_TLSV1_1
-        | SslOptions::NO_TLSV1_2
-        | SslOptions::NO_TLSV1_3;
+        | SslOptions::NO_TLSV1_2;
 
     ctx.clear_options(no_ssl_mask);
     let mut options = SslOptions::empty();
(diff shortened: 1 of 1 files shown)


src/imp/openssl.rs
```diff
@@ -54,7 +55,8 @@ fn supported_protocols(
         | SslOptions::NO_SSLV3
         | SslOptions::NO_TLSV1
         | SslOptions::NO_TLSV1_1
-        | SslOptions::NO_TLSV1_2;
+        | SslOptions::NO_TLSV1_2
+        | SslOptions::NO_TLSV1_3;
```
sfackler, line 59: OpenSSL versions too old for set_min_proto_version won't have TLSV1_3 support afaik.

sfackler:
What about Security.framework?

sfackler:
Bumping the Rust version used in the CI builds should unbreak those.

src/imp/openssl.rs
RyanMeulenkamp: I'm not sure what you mean. The OpenSSL versions that don't have min protocol config are very old and no longer supported I think (< v1.1.0), while TLS1.3 was introduced at OpenSSL v1.1.1, also pretty old by now.

sfackler: This code is only used for old versions of OpenSSL that won't have a `NO_TLSV1_3` symbol.

RyanMeulenkamp: I see. Didn't notice this is inside a `#[cfg()]` block.

src/imp/openssl.rs
```diff
@@ -32,6 +32,7 @@ fn supported_protocols(
             Protocol::Tlsv10 => SslVersion::TLS1,
             Protocol::Tlsv11 => SslVersion::TLS1_1,
             Protocol::Tlsv12 => SslVersion::TLS1_2,
+            Protocol::Tlsv13 => SslVersion::TLS1_3,
```
sfackler, line 35: I believe this will break compilation against OpenSSL 1.1.0, which has set_min_proto_version but does not have TLS 1.3 support. You'll need to add some extra version logic in the build script and here.

TheWaWaR:
Any update on this?

src/imp/openssl.rs
Sytten: What about a feature flag? Feels like this could get merged quicker by letting the users decide

axelsimon:
Is this blocked on putting it behind a feature flag, to maintain compatibility with OpenSSL 1.1.0, or is something else keeping it stuck?

Neustradamus:
Linked to:
- https://github.com/rust-native-tls/rust-native-tls/pull/353
- https://github.com/rust-native-tls/rust-native-tls/issues/140
- https://github.com/rust-native-tls/rust-native-tls/pull/159
- https://github.com/rust-native-tls/rust-native-tls/pull/291
- https://github.com/rust-native-tls/rust-native-tls/pull/235
- https://github.com/rust-native-tls/rust-native-tls/pull/278
- https://github.com/retep998/winapi-rs/pull/859
- https://github.com/retep998/winapi-rs/pull/963
- https://github.com/element-hq/element-x-ios/issues/786
- https://github.com/matrix-org/matrix-rust-sdk/pull/6053
- https://github.com/deltachat/deltachat-ios/issues/1693

PR #249, opened 2023-02-25
Vuln Patch: Update `tempfile` to `v3.4.0`

kevinpark1217:
There is a security vulnerability in the dependency of the `tempfile` which was fixed in `v3.4.0` by removing the vulnerable dependency.

Related: https://github.com/Stebalien/tempfile/pull/208


kpark-hrp:
@sfackler Hey, it would be great if this can be prioritized.

And if you can release a patch version after the merge, that would be awesome.

sfackler:
This is not a necessary change. Just run `cargo update`.

PR #250, opened 2023-02-28
Update to tempfile 3.4.0

quad:
The newer version of tempfile drops its dependency on the version of remove_dir_all with a recently announced vulnerability: https://github.com/advisories/GHSA-mc8h-8q98-g5hr


sfackler:
https://github.com/sfackler/rust-native-tls/pull/249#issuecomment-1445219493

PR #259, opened 2023-04-20
Remove unneeded checks for private key

UXabre:
Solves: https://github.com/sfackler/rust-native-tls/issues/238

THe problem is that EC keys slightly change the separator text and I don't think a check is needed on this level... what if, in the future, a different thing is added? 

I think this is part of the appeal of using native TLS, that it can support things not originally supported. That is why I propose to remove the check altogether

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+## [v0.2.12]
+
+### Fixed
+
+* Remove unneeded private key check for PKCS#8
+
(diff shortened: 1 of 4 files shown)


sfackler:
You probably have an EC key in Sec1 format, not pkcs#8. The OpenSSL cli can help you convert your private key file to the appropriate format.

UXabre:
Ok, now i feel stupid for not realizing this. Thanks for taking the time to reply!

PR #261, opened 2023-05-04
Add implementation for mbedtls

### LINKED ISSUE #211: mbedtls implementation
Without `rust-native-tls` support, a lot of std crates become partially, or completely unusable. For our std target, `riscv32imc-esp-espidf`, we have an mbedtls implementation available.

[`rust-mbedtls`](https://github.com/fortanix/rust-mbedtls) is a stable and well maintained crate with bindings to mbedtls. Would you be open to PR's adding support in `rust-native-tls`?

zvolin:
Yo,
this is my attempt to resolve #211. The implementation is based on the initial work done by @MabezDev.

There are a few things worth mentioning, where I'd like to use some guidance:
- this breaks api compatibility. I've changed `fn get_ref(&self) -> &S` and `fn get_mut(&mut self) -> &mut S`  to `fn get_ref(&self) -> impl Deref<Target = S>` and `fn get_mut(&mut self) -> impl DerefMut<Target = S>`. I'm not sure if that's required, I started with a bit different implementation, but it's the non-unsafe way to have `TlsStream` impl `Sync`. I've checked 3 top dependant crates: `Reqwest`, `hyper-tls` and `tokio-native-tls` and only the last one failed, on tests of `smoke.rs`. If we decide to go this way I'm happy to run tests on all dependant crates and create the issues that'll warn about bumping this. If not, I can try the other ways. The other possible solutions would be to make this change only on mbedtls targets (but not uniform api is a painpoint) / try to implement this differently (but likely with unsafe).
- I'm not sure what would be the way to get 'default' ca root certs list. My use case are Intel SGX enclaves which do not have (simple) access to the filesystem. I know there is interest in running those on some esp platforms etc, but I'm unfamiliar with them. We could maybe hardcode the ca root certificates from Mozilla but I decided that for now I'll start with the mindset that the user needs to provide them himself in case of using `mbedtls`.
- What other targets to add? I will add the ones mentioned in the issue, are there maybe more 'obvious ones'?
- Should I include builds for those targets in the CI? Will require installing additional toolchains and can extend the overall time / usage

A disclaimer, I'm not really familiar with HTTPS/TLS/SSL spec and implementations and also with networking in general :sweat_smile: so I'll be thankful for a careful review


src/imp/mbedtls.rs
```diff
@@ -0,0 +1,575 @@
...
+        // the only single mutable reference rule?
+        // But this shouldn't be possible without unsafe
+        MutexGuard::map(self.ctx.lock(), |ctx| {
+            ctx.io_mut()
+                .expect("Not connected")
+                .downcast_mut::<S>()
+                // unwrap is safe as this is the exact stream passed to context
+                .unwrap()
+        })
+    }
```
zvolin, line 267: Calling this function now may easily deadlock, if any other method (eg. this itself) that locks the context is called while holding the reference. this could be a `ReentrantMutex` instead, but not yet sure if we can get `get_mut` working then (lock doesn't provide `&mut`). Will try with `RefCell` in the meantime but again, it's not that simple :sweat_smile:

Cargo.toml
```diff
@@ -26,12 +26,22 @@ tempfile = "3.1.0"
...
 
-[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))'.dependencies]
+[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios", target_env = "sgx")))'.dependencies]
 log = "0.4.5"
 openssl = "0.10.29"
 openssl-sys = "0.9.55"
 openssl-probe = "0.1"
 
+[target.'cfg(target_env = "sgx")'.dependencies]
+mbedtls = { version = "0.8.1", features = ["pkcs12", "pkcs12_rc2", "std", "rdrand", "mpi_force_c_code" ], default-features = false }
```
sfackler, line 36: mbedtls 0.9 is out, which looks like it should simplify a lot of stuff in the implementation.

src/imp/mbedtls.rs
```diff
@@ -0,0 +1,575 @@
...
+impl Debug for Certificate {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Certificate")
+            .field(&self.0.as_der())
+            .finish()
+    }
+}
+
+pub struct TlsStream<S: Stream> {
+    ctx: Arc<Mutex<Context>>,
```
sfackler, line 247: What's the purpose of the arc and mutex?

zvolin:
```rust
    is_sync::<TlsStream<TcpStream>>();
    is_send::<TlsStream<TcpStream>>();
```

Cargo.toml
zvolin: yeah, I know, however up to my knowledge `fortanix` itself hasn't yet [adopted it](https://github.com/fortanix/rust-sgx/blob/master/em-app/Cargo.toml#L17) in their `rust-sgx` code. This is only my personal requirement tho, and thus it doesn't matter much so if you want me to change it I'll do :+1:

zvolin: the more I think about it the more I'm not sure if that changes anything. Tomorrow I'll check if it's fine to have 2 different versions linked to the same crate and if so, I'll try the bump

src/imp/mbedtls.rs
```diff
@@ -0,0 +1,575 @@
...
+        use std::ffi::CStr;
+
+        let ctx_lock = self.ctx.lock();
+        let raw_ctx: *const mbedtls_sys::ssl_context = (&*ctx_lock).into();
+        let alpn = unsafe { CStr::from_ptr((*raw_ctx).alpn_chosen).to_bytes().to_vec() };
+        if alpn.is_empty() {
+            Ok(Some(alpn))
+        } else {
+            Ok(None)
+        }
```
zvolin, line 294: the condition should probably be negated, but need to check why the tests pass then :detective:

src/imp/mbedtls.rs
sfackler: I would just flag those tests off when building against mbedtls.

Cargo.toml
zvolin: It's indeed not a problem to use an old and new version in the same project, so I'll try to switch to the `0.9`

zvolin: Updated to `0.9`, indeed simplified a lot

src/imp/mbedtls.rs
zvolin: no longer relevant

src/imp/mbedtls.rs
zvolin: Disabled them and removed the arc and mutex. Will need to check if that is fine when using crates depending on this

src/imp/mbedtls.rs
zvolin: fixed and added them to ci

src/imp/mbedtls.rs
zvolin: It looks like at least `ureq` depends on `TlsStream` being `Sync`. `reqwest` and `hyper-tls` doesn't build for my target anyway as they depend on `socket2`, so it's harder to check those

jethrogb: All platforms should offer the same Send/Sync guarantees.

jethrogb (commented):
Let's investigate if MbedTLS can have more things Sync.

Cargo.toml
```diff
@@ -26,12 +26,21 @@ tempfile = "3.1.0"
...
-[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios")))'.dependencies]
+[target.'cfg(not(any(target_os = "windows", target_os = "macos", target_os = "ios", target_env = "sgx")))'.dependencies]
 log = "0.4.5"
 openssl = "0.10.29"
 openssl-sys = "0.9.55"
 openssl-probe = "0.1"
 
+[target.'cfg(target_env = "sgx")'.dependencies]
+mbedtls = { version = "0.9", features = ["std", "rdrand", "mpi_force_c_code" ], default-features = false }
+mbedtls-sys-auto = { version = "2.28.0", default-features = false }
```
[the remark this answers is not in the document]
zvolin: removed

src/imp/mbedtls.rs
```diff
@@ -0,0 +1,522 @@
...
+        // and construct pkcs5 decoder from it
+        let scheme = pkcs5::EncryptionScheme::try_from(&algorithm_der[..])?;
+
+        Ok(scheme.decrypt(pass.as_ref(), &key_bag.encrypted_data)?)
+    } else {
+        Err(Error::Custom(
+            "Unsupported key encryption algorithm".to_owned(),
+        ))
+    }
+}
```
zvolin, line 126: Custom pkcs12 treatment is there only because `rust-mbedtls` implementation doesn't support pbes2 encrypted keys (which is openssl's default). I'll try to provide support for it directly in mbedtls instead

zvolin: @sfackler @jethrogb
I've checked and we could add an implementation of pbes2 to rust-mbedtls, however from encryption schemes defined by [pkcs5 rfc](https://datatracker.ietf.org/doc/html/rfc8018#appendix-B.2), MbedTls [has support](https://github.com/fortanix/rust-mbedtls/blob/master/mbedtls-sys/vendor/library/oid.c#L550) only for des-cbc and des-ede3-cbc.
rc2-cbc, rc5-cbc and aes-cbc are missing and aes-cbc is openssl's default when running `openssl pkcs12 -export`.
rc2 and rc5 are also missing in `pkcs5` crate, so it seems to be all about AES.
Moreover, des-cbc is considered insecure: 'Although its short key length of 56 bits makes it too insecure for modern applications' [wiki](https://en.wikipedia.org/wiki/Data_Encryption_Standard)

Here are the options:
 - Add native (MbedTls) support for pbes2 in `rust-mbedtls` and use it there, only allowing des-cbc and des-ede3-cbc encryption schemes and providing compatible `.p12` for `rust-native-tls` tests
 - Add non-native (using pkcs5 crate) in `rust-mbedtls` and maybe hide it behind feature flag, allowing also aes-cbc
 - Keep this implementation and don't use `rust-mbedtls` for pbes2 (however if I'll have some spare time I'll likely issue a PR for `rust-mbedtls` anyway as I've already written most of the implementation)

src/imp/mbedtls.rs
zvolin: As fortanix/rust-mbedtls/pull/262 is merged, I'll wait for release of `0.9.1` and add `Sync` support :ok_hand:

src/imp/mbedtls.rs
zvolin: https://github.com/Mbed-TLS/mbedtls/issues/7024
https://github.com/Mbed-TLS/mbedtls/issues/7038
It's possible AES will get support in MbedTls at some point, but it'll probably require rust-mbedtls to update to 3.x then, which can be major effort

src/imp/mbedtls.rs
zvolin: done

src/imp/mbedtls.rs
jethrogb: > it'll probably require rust-mbedtls to update to 3.x

This is actually ongoing

zvolin: The preferred way of solving this would be to add support for AES in `MbedTls`: https://github.com/Mbed-TLS/mbedtls/pull/7604
then pull it into `rust-mbedtls` and use it directly here

jethrogb:
@sfackler you wrote

>  I'd probably just recommend people not use PKCS#12 anymore.

https://github.com/sfackler/rust-native-tls/issues/245#issuecomment-1324355197

Can we deprecate the p12 API and not require it for new platform implementations?

src/imp/mbedtls.rs
zvolin: we've been discussing this with @jethrogb and it seems that `rust-mbedtls` is rather going to take steps towards removing pkcs12 support than extending it to pbes2. So it is up to you @sfackler whether we leave this implemented as is now using p12 and pkcs5 crates or just mark it as unimplemented.

kornelski:
I've rebased it and updated:

https://github.com/rust-native-tls/rust-native-tls/tree/zvolin-feature/mbedtls-imp

jethrogb:
Note that the [rust-mbedtls crate is deprecated](https://github.com/fortanix/rust-mbedtls/?tab=readme-ov-file#change-of-status) due to MbedTLS 3.x being no longer thread safe. The `x86_64-fortanix-unknown-sgx` target maintainers no longer recommend defining MbedTLS as “native” on this target.

PR #269, opened 2023-06-21
update to openssl 0.10.55 to fix vulnerability report

Freyskeyd:
Update `openssl` to `0.10.55` to fix security vulnerability:

- https://rustsec.org/advisories/RUSTSEC-2023-0044


sfackler:
Run `cargo update`.

Freyskeyd:
@sfackler yes, but shouldn't we restrict the dep?

sfackler:
Why?

PR #279, opened 2023-09-15
Remove `once_cell`

tamird:
Mutex::new is const since 1.63.0, and this library seems to have MSRV
1.63.0 since https://github.com/sfackler/rust-native-tls/pull/254.


sfackler:
Build's red

tamird:
Should be fixed now.

PR #280, opened 2023-10-11
Use PfxImportOptions::named_no_persist_key()

ironyman:
Use the new method PfxImportOptions::named_no_persist_key() from [schannel-rs](https://github.com/steffengy/schannel-rs/pull/97) to prevent leaking keys when importing pkcs12 files.


kornelski:
this needs a fix

PR #282, opened 2023-11-06
Support exporting pem from `Certificate`

szclsya:
Nothing fancy. Add support to generate PEM formatted certificate from `Certificate`, in addition to the already-supported DER format.

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -201,6 +201,11 @@ impl Certificate {
         let der = self.0.to_der()?;
         Ok(der)
     }
+
+    pub fn to_pem(&self) -> Result<Vec<u8>, Error> {
+        let pem = self.0.to_pem()?;
+        Ok(pem)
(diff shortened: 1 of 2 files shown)


src/imp/security_framework.rs
```diff
@@ -225,6 +225,10 @@ impl Certificate {
     pub fn to_der(&self) -> Result<Vec<u8>, Error> {
         Ok(self.0.to_der())
     }
+
+    pub fn to_pem(&self) -> Result<Vec<u8>, Error> {
+        panic!("Not implemented on macOS/iOS");
```
sfackler, line 230: This'll need to be implemented.

szclsya: Seems that there's currently no implementation in the underlying [rust-security-framework](https://github.com/kornelski/rust-security-framework) crate. I'll see if I can do anything there.

PR #293, opened 2024-04-02
Set default minimum protocol to TLS 1.2

Property404:
TLS 1.0 (published 1999) and 1.1 (published 2006) have been deprecated since 2021 ([RFC 8996](https://datatracker.ietf.org/doc/html/rfc8996)) and are no longer considered secure.

I think this could be considered a breaking change


sfackler:
I think it's probably best to just make the change rather than cutting a major break. I expect most of the TLS backends people are using under native-tls have already turned off support for 1.0 and 1.1.

PR #300, opened 2024-07-23
bump openssl to fix RUSTSEC-2024-0357

stormshield-gt:
This MR bump openssl crate to include the fix for [RUSTSEC-2024-0357](https://rustsec.org/advisories/RUSTSEC-2024-0357.html)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ schannel = "0.1.17"
 
 [target.'cfg(not(any(target_os = "windows", target_vendor = "apple")))'.dependencies]
 log = "0.4.5"
-openssl = "0.10.29"
+openssl = "0.10.66"
 openssl-sys = "0.9.55"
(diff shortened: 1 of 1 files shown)


sfackler:
This is not necessary. Run cargo update locally.

stormshield-gt:
Just to understand, by not bumping the version here, `cargo` is allowed to resove `openssl` to a lower version that contain the vulnerability right?

We don't want to prevent this by only allowing version `>= 0.10.66`  ?

sfackler:
That's correct.

stormshield-gt:
Why don't we want to prevent this ? Because the vulnerability is not critical enough ? Or maybe we don't want the churn of everybody recompiling if they don't care about the vulnerability ?

stormshield-gt:
Oh I see, thanks a lot for taking the time to answer my question

PR #314, opened 2025-01-17
fix(openssl.rs): replaced pkcs12.parse2() with pkcs12.parse()

XlebyllleK:
```rust
error[E0599]: no method named `parse2` found for struct `Pkcs12` in the current scope
   --> /home/celestifyx/.cargo/registry/src/index.crates.io-6f17d22bba15001f/native-tls-0.2.12/src/imp/openssl.rs:162:29
    |
162 |         let parsed = pkcs12.parse2(pass)?;
    |                             ^^^^^^
    |
help: there is a method `parse` with a similar name
    |
162 |         let parsed = pkcs12.parse(pass)?;
    |                             ~~~~~
```

### DIFF
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -159,7 +159,7 @@ pub struct Identity {
 impl Identity {
     pub fn from_pkcs12(buf: &[u8], pass: &str) -> Result<Identity, Error> {
         let pkcs12 = Pkcs12::from_der(buf)?;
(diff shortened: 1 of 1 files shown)


sfackler:
The correct fix was in https://github.com/sfackler/rust-native-tls/commit/2424bc5efd1b8b4bcf60dbda93259a3f29db7f06, which just needs to be published.

PR #318, opened 2025-02-05
update openssl due to RUSTSEC-2025-0004

CramBL:
Simply bumps openssl to get the fix for https://rustsec.org/advisories/RUSTSEC-2025-0004


sfackler:

PR #322, opened 2025-02-13
update openssl to 0.10.70

pvichivanives:
Hi,
I saw that open-ssl had a security vulnerability here  https://rustsec.org/advisories/RUSTSEC-2025-0004.html
, which is being reported by Rust Audit.

Just wanted to raise a PR to bump the version:


sfackler:

PR #329, opened 2025-04-18
fix: cargo audit warning over openssl version

frederictaillandier:
updating to 0.10.72 as recommended

https://rustsec.org/advisories/RUSTSEC-2025-0022


sfackler:

frederictaillandier:
thanks !

PR #294, opened 2024-04-09
Added visionos target support

simlay:
Apple Vision Pro support was added as a tier 3 target to rust in https://github.com/rust-lang/rust/pull/121419 with `target_os` of `visionos`.

Depends on:
- [ ] https://github.com/kornelski/rust-security-framework/pull/201
- [x] https://github.com/rust-lang/libc/pull/3568 to actually build the stdlib with the `-Zbuild-std` flag for `aarch64-apple-visionos` or the simulator target.

PR #330, opened 2025-04-19
Use `#[cfg(target_os = "macos")]`

madsmtm:
Instead of the very verbose:
```rust
#[cfg(not(any(
   target_os = "ios",
   target_os = "watchos",
   target_os = "tvos",
   target_os = "visionos"
)))]
```

I suspect this is fine, as the macOS functionality is implemented using deprecated functions, and thus likely to also be unavailable if Apple decides to add a future platform.

CC @simlay.

### DIFF
--- a/src/imp/security_framework.rs
+++ b/src/imp/security_framework.rs
@@ -18,54 +18,24 @@ use std::str;
 use std::sync::Mutex;
 use std::sync::Once;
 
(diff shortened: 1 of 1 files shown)


simlay (approved):
This is definitely cleaner than what I had done in #294.

PR #334, opened 2025-05-04
chore: Bump openssl and openssl-sys

### LINKED ISSUE #333: Bump openssl to 0.10.72 to avoid the UAF bugs
openssl before 0.10.72 contains two use-after-free bugs.

https://github.com/sfackler/rust-openssl/commit/d1a12e21573e95727b2e38b8b65273cb389be7e4

Even though you can download the latest version (0.10.72) from the official registry crates.io for the current build,
as of May 4, you may still download the 0.10.71 version from [rsproxy sparse](https://rsproxy.cn/), which is one of the most used mirror registries in China.

Therefore I updated the openssl and openssl-sys versions in Cargo.toml and force cargo to download the 0.10.72 version.

Chain-Fox:
Fix #333 

Update the openssl and openssl-sys versions in Cargo.toml,
Thus force cargo to download the latest version of openssl even from a sparse mirror registry ( which may auto download an older version).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,8 +29,8 @@ schannel = "0.1.17"
 
 [target.'cfg(not(any(target_os = "windows", target_vendor = "apple")))'.dependencies]
 log = "0.4.5"
-openssl = "0.10.69"
-openssl-sys = "0.9.81"
+openssl = "0.10.72"
+openssl-sys = "0.9.108"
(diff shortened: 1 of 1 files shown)


sfackler:
Why would your sparse registry find the new version of this crate and not the new version of openssl and openssl-sys?

PR #371, opened 2026-02-18
Fix/openssl tlsv13 match 0.2.18

Kolzodev712:
Fixes the OpenSSL backend compile failure on Linux (and other OpenSSL targets) where Protocol::Tlsv13 was not handled in the legacy supported_protocols path, causing “non-exhaustive patterns” at the min/max protocol matches. 
Adds the missing Tlsv13 arms in src/imp/openssl.rs, bumps the version to 0.2.18, and documents the fix in the changelog.


tronicboy1:
This is a pretty big mistake in code review. I wonder if there is anything we can do to ensure this doesn't happen again?

kornelski:
It was a combination of two pitfalls: 

1. the CI doesn't test with old openssl versions
2. testing git checkout isn't exactly the same as testing published package

I've added CI tests for the second case. I'm adding more CI tests for various platforms to cover more openssl flavors.

PR #376, opened 2026-03-11
Bump schannel to 0.1.29

nateprewitt:
This PR updates schannel-rs to 0.1.29 which released yesterday. Prior to this release, a bug in schannel-rs could result in TLS record bytes leaking into application data when using TLS 1.3 on Windows, This leads to silent data corruption while sending large payloads. See https://github.com/steffengy/schannel-rs/pull/121 for full details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ libc = "0.2.170"
 tempfile = "3.1.0"
 
 [target.'cfg(target_os = "windows")'.dependencies]
-schannel = "0.1.28"
+schannel = "0.1.29"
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hi @kornelski, I wanted to soft ping this. No rush, but it does pose a notable integrity risk for anyone using native-tls on Windows. I just want to make sure it's on your radar. Thanks for your work on the project!

sfackler:
This is not necessary. Run `cargo update`.

nateprewitt:
Ok, fair enough. We've already handled this in our projects. It's just coming through Reqwest still when using `-Z minimal-versions`. If that's not a concern for the project, I'll close this out. Thanks for the response!

PR #377, opened 2026-05-13
Add TLS exporter (RFC 5705 / RFC 8446 §7.5) support

Ylarod:
Adds `TlsStream::export_keying_material`, exposing the TLS exporter
([RFC 5705] for TLS 1.0–1.2, [RFC 8446] §7.5 for TLS 1.3) so callers can
derive exported keying material for uses like [RFC 9266] `tls-exporter`
channel binding and SRTP key derivation.

- **OpenSSL backend**: wires through `SSL_export_keying_material`; works on
  TLS 1.2 and TLS 1.3, honoring the RFC 5705 distinction between
  `context = None` and `context = Some(&[])`.
- **SChannel / Secure Transport backends**: the underlying Rust bindings
  do not surface this primitive, so the method returns an unsupported-style
  error (`io::ErrorKind::Unsupported` on Windows, `errSecUnimplemented` on
  Apple) rather than silently producing nothing. This keeps the API uniform
  across platforms while making the limitation explicit.
- Adds round-trip tests asserting that two peers derive the same EKM and
  that different labels/contexts produce different output.
- Documents backend coverage on the public method and adds a CHANGELOG entry.

Also includes a small drive-by: switches an `ok_or_else(|| Error::EmptyChain)`
to `ok_or(Error::EmptyChain)` to silence `clippy::unnecessary_lazy_evaluations`
on Rust 1.95+ (the closure was redundant because the variant is a plain
constant).

[RFC 5705]: https://tools.ietf.org/html/rfc5705
[RFC 8446]: https://tools.ietf.org/html/rfc8446
[RFC 9266]: https://tools.ietf.org/html/rfc9266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## Unreleased
+
+* Added `TlsStream::export_keying_material` for deriving exported keying
+  material (RFC 5705 / RFC 8446 §7.5), as used by RFC 9266 channel binding.
+  Supported on the OpenSSL backend; returns an error on SChannel (Windows)
+  and Secure Transport (Apple) where the underlying bindings do not expose
+  the primitive.
+
 ## [v0.2.18]
 
 * Fixed min/max protocol selection fallback for very old OpenSSL versions.
--- a/src/imp/openssl.rs
+++ b/src/imp/openssl.rs
@@ -120,6 +120,7 @@ pub enum Error {
     EmptyChain,
     NotPkcs8,
     AlpnTooLong,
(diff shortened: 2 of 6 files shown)


kornelski:
`native-tls` is a cross-platform abstraction. Every feature that doesn't work on all platforms makes it fail at its primary goal :(

Apple's Security framework is an outdated abandonware, which drags everything down, so it creates a precedent for moving past it with functionality.

But I'm worried that this way `native-tls` will end up being just the `openssl` crate with extra steps, by slowly copypasting every function from it.

